PR #46, opened 2024-10-10
Feature/serialization

luuk234:
[ADDED} scene serialization
[ADDED] save button
[ADDED] entity serialization interface


MuskettaMan (changes requested):

santiberna (changes requested):
I think you could move the Entity Serializer into another (header probably) file, but otherwise I think the question of saving with cereal or filepath is not really important unless we run into limitations

MuskettaMan (approved):
Looks good, approved, left some minor feedback on cleaning up the source files

santiberna (approved):
Okay looks a lot better. You can merge!

PR #62, opened 2024-10-28
Physics update

dumitrurares14:
### Description

1)

- Added more control over the rigid bodies
- Rigid bodies can now be created with different shapes (box and spheres)
- the editor can now modify position, rotation, scale, collision type  
image

2)
- Added ray casts to the physics system that return everything we need in order to properly use this for development later.
```cpp
struct RayHitInfo
{
    entt::entity entity = entt::null; // Entity that was hit
    glm::vec3 position = glm::vec3(0.0f, 0.0f, 0.0f); // Position where the ray hits; HitPoint = Start + mFraction * (End - Start)
    float hitFraction = 0.0f; // Hit fraction of the ray/object [0, 1], HitPoint = Start + mFraction * (End - Start)
    bool hasHit = false;
};
```
- A basic persistent line visualizer is implemented for the rays using the existing debug pipeline. This will be reworked when we finish implementing the proper debug line data structure

3)
Physics module is actually a module now.

### Issues

https://bubonic-brotherhood.codecks.io/card/16q-physics-system-basic-colliders
https://bubonic-brotherhood.codecks.io/card/194-ray-casts
https://bubonic-brotherhood.codecks.io/decks/4-character/card/18v-shooting-user-story
https://bubonic-brotherhood.codecks.io/decks/7-engine/card/1cj-make-the-physics-system-a-module-as-i-have-never-used-the-cmake-modules

### Test criteria

- [ ] Spawn a few objects and play around with the editor, let me know if everything works as you would expect.
- [ ] Shoot a ray by pressing space towards the camera forward direction. The hit info should be printed to the console.


engine/include/systems/physics_system.hpp
```diff
@@ -10,16 +19,21 @@ class PhysicsSystem : public System
     PhysicsSystem(ECS& ecs, PhysicsModule& physicsModule);
     ~PhysicsSystem() = default;
 
-    entt::entity CreatePhysicsEntity();
-    void CreatePhysicsEntity(RigidbodyComponent& rb);
     void AddRigidBody(entt::entity entity, RigidbodyComponent& rigidbody);
+    [[nodiscard]] RayHitInfo ShootRay(const glm::vec3& origin, const glm::vec3& direction, float distance) const;
```
luuk234, line 23: we have NO_DISCARD macro for this now

engine/include/components/rigidbody_component.hpp
```diff
@@ -1,19 +1,47 @@
 ﻿#pragma once
-#include "modules/physics_module.hpp"
+#include "physics_module.hpp"
 
 struct RigidbodyComponent
 {
     // default creates a sphere at 0,2,0
-    RigidbodyComponent(JPH::BodyInterface& bodyInterface)
+    RigidbodyComponent(JPH::BodyInterface& bodyInterface, PhysicsShapes shape = eSPHERE, entt::entity ownerEntity = entt::null)
```
santiberna, line 7: This seems incorrect, the owner entity should never be entt::null, why is it a default parameter on the constructor? (seems error prone to me)

engine/include/systems/physics_system.hpp
```diff
@@ -10,16 +19,21 @@ class PhysicsSystem : public System
...
     void CleanUp();
 
     void Update([[maybe_unused]] ECS& ecs, [[maybe_unused]] float deltaTime) override;
     void Render([[maybe_unused]] const ECS& ecs) const override;
     void Inspect() override;
+    void InspectRigidBody(RigidbodyComponent& rb);
+
+    void SetCameraPosition(const glm::vec3& position) { _cameraPosition = position; }
```
luuk234, line 31: what does the camera inside the physics system do?

engine/include/components/rigidbody_component.hpp
```diff
@@ -1,19 +1,47 @@
...
+        shapeType = shape;
+        if (shape == eSPHERE)
+        {
+            JPH::BodyCreationSettings const bodySettings(new JPH::SphereShape(0.5f), JPH::Vec3(0.0, 2.0, 0.0), JPH::Quat::sIdentity(), JPH::EMotionType::Dynamic, PhysicsLayers::MOVING);
+            bodyID = bodyInterface.CreateAndAddBody(bodySettings, JPH::EActivation::Activate);
+        }
+        else if (shape == eBOX)
+        {
+            JPH::BodyCreationSettings const bodySettings(new JPH::BoxShape(JPH::Vec3(0.5, 0.5, 0.5)), JPH::Vec3(0.0, 2.0, 0.0), JPH::Quat::sIdentity(), JPH::EMotionType::Dynamic, PhysicsLayers::MOVING);
+            bodyID = bodyInterface.CreateAndAddBody(bodySettings, JPH::EActivation::Activate);
```
luuk234, line 18: doesnt cover eCONVEX_HULL case, if we don't support this should remove it from the enum or throw error/message

engine/physics/public/physics_module.hpp
```diff
@@ -211,13 +230,19 @@ class DebugRendererSimpleImpl : public JPH::DebugRendererSimple
         return linePositions;
     }
 
+    [[nodiscard]] const std::vector<glm::vec3>& GetPersistentLinesData() const
```
luuk234, line 233: NO_DISCARD

engine/include/systems/physics_system.hpp
```diff
@@ -10,16 +19,21 @@ class PhysicsSystem : public System
     PhysicsSystem(ECS& ecs, PhysicsModule& physicsModule);
     ~PhysicsSystem() = default;
```
santiberna, line 20: Destructor should be declared with override (its a subclass of ModuleInterface afterall)

engine/physics/public/physics_module.hpp
dumitrurares14: have we actually decided on this? there are places where [[nodiscard]] is still being used

engine/source/engine.cpp
```diff
@@ -192,6 +194,18 @@ void OldEngine::Tick(Engine& engine)
...
         JPH::RVec3Arg cameraPos = { _scene->camera.position.x, _scene->camera.position.y, _scene->camera.position.z };
         _physicsModule->debugRenderer->SetCameraPos(cameraPos);
+
+        // shoot rays
+        if (ImGui::IsKeyPressed(ImGuiKey_Space))
+        {
+            const glm::vec3 cameraDir = (glm::quat(_scene->camera.eulerRotation) * -FORWARD);
+            const RayHitInfo hitInfo = _ecs->GetSystem<PhysicsSystem>().ShootRay(_scene->camera.position + glm::vec3(0.0001), glm::normalize(cameraDir), 5.0);
+
+            std::cout << "Hit: " << hitInfo.hasHit << std::endl
```
luuk234, line 204: bblog if this is meant to stay

luuk234 (changes requested):
minor stuff, looks very solid overral

VengarvdWal (commented):
Luuk makes some good points, I was also not aware of the NO_DEBUG macro. Also the persisent line data is a hacky implementation but we discussed that this will be refactored later

engine/include/systems/physics_system.hpp
santiberna, line 23: Is there a reason this function is not a part of the Physics Module? nothing related to the ECS is here except the return value containing an entt::entity (which can be in the physics module as well)

engine/include/systems/physics_system.hpp
```diff
@@ -10,16 +19,21 @@ class PhysicsSystem : public System
...
     void Update([[maybe_unused]] ECS& ecs, [[maybe_unused]] float deltaTime) override;
     void Render([[maybe_unused]] const ECS& ecs) const override;
     void Inspect() override;
+    void InspectRigidBody(RigidbodyComponent& rb);
+
+    void SetCameraPosition(const glm::vec3& position) { _cameraPosition = position; }
+    void SetCameraDirection(const glm::vec3& direction) { _cameraDirection = direction; }
 
 private:
     ECS& _ecs;
```
santiberna, line 35: Also really avoid references in data members (for future reference, I know it is hard to change now).

engine/source/systems/physics_system.cpp
```diff
@@ -1,31 +1,46 @@
 ﻿#include "systems/physics_system.hpp"
+#include "components/name_component.hpp"
 #include "components/rigidbody_component.hpp"
 #include "ecs.hpp"
 #include "imgui/imgui.h"
-#include "modules/physics_module.hpp"
+#include "physics_module.hpp"
 
 PhysicsSystem::PhysicsSystem(ECS& ecs, PhysicsModule& physicsModule)
```
santiberna, line 8: IMO, I think the PhysicsSystem should also be a part of the physics module (not sure how this would integrate with the OldEngine)

engine/physics/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_library(Physics)
+module_default_init(Physics)
+
+target_link_libraries(Physics
+    PUBLIC Core
+    PUBLIC Application
+    PUBLIC OldEngine
```
santiberna, line 7: Do you really need access to OldEngine? Which dependencies do you have on OldEngine?

santiberna (changes requested):
In general, the implementation is correct and the work for debug line drawing and ImGui is good, but there are some lacking aspects in your code that you should see in comments. I also think you should:
- Look at the Linter errors: there are issues about uninitialized variables and unformatted files that you can trivially fix
- The linter also notes some missing includes (not caught in build checks because of include order saving you)
- In general, I think other physics related stuff like the RigidBody component and Physics System should be in the PhysicsModule

engine/include/systems/physics_system.hpp
dumitrurares14: My bad, i forgot it there when i was trying something

engine/physics/public/physics_module.hpp
luuk234: If i remember coorectly we have, those cases should be resolved then when we get to it as well.

engine/include/systems/physics_system.hpp
dumitrurares14: I moved it now to physics module 🫡

engine/physics/CMakeLists.txt
dumitrurares14: No, it was a mistake. I updated it now

santiberna (approved):
Good Job!

luuk234 (approved):

PR #65, opened 2024-10-29
Feature/demo

MuskettaMan:
### Description

Added some demo assets into our world and resolved some small issues with shadows and PBR. PBR still looks broken on flat surfaces, which has to be resolved later.


engine/renderer/private/gbuffers.cpp
```diff
@@ -15,10 +15,15 @@ GBuffers::GBuffers(const std::shared_ptr<GraphicsContext>& context, glm::uvec2 s
...
-    _shadowFormat = supportedDepthFormat.value();
+    if (supportedDepthFormat.has_value())
+    {
+        _depthFormat = supportedDepthFormat.value();
+        _shadowFormat = supportedDepthFormat.value();
+    }
+    else
+    {
+        assert(false && "No supported depth format!");
+    }
```
mmzala, line 26: Why change this to an if statement when the assert exits the program anyway when there is no value available? Couldn't this just have been left as previously, like this:
`assert(supportedDepthFormat.has_value() && "No supported depth format!");`

engine/renderer/private/pipelines/skydome_pipeline.cpp
```diff
@@ -54,7 +54,8 @@ void SkydomePipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t c
     colorAttachmentInfos[0].imageView = _context->Resources()->ImageResourceManager().Access(_hdrTarget)->views[0];
     colorAttachmentInfos[0].imageLayout = vk::ImageLayout::eAttachmentOptimalKHR;
     colorAttachmentInfos[0].storeOp = vk::AttachmentStoreOp::eStore;
-    colorAttachmentInfos[0].loadOp = vk::AttachmentLoadOp::eLoad;
+    colorAttachmentInfos[0].loadOp = vk::AttachmentLoadOp::eLoad; // eClear;
+    colorAttachmentInfos[0].clearValue.color = vk::ClearColorValue { .float32 = { { 0.0f, 0.0f, 0.0f, 0.0f } } };
```
mmzala, line 58: Clear color is not needed

engine/source/pipelines/skydome_pipeline.cpp
mmzala, line 1: File should be removed

engine/renderer/private/pipelines/geometry_pipeline.cpp
```diff
@@ -1,5 +1,6 @@
 #include "pipelines/geometry_pipeline.hpp"
 
+#include "../vulkan_helper.hpp"
```
mmzala, line 3: Can be removed as it is already included at the bottom of includes

external/imgui_entt_entity_editor/imgui_entt_entity_editor.hpp
```diff
@@ -1,9 +1,9 @@
 // for the license, see the end of the file
 #pragma once
 
+#include <functional>
```
mmzala, line 4: Was this an automatic change or an accident? The change is not really needed

mmzala (changes requested):
Nice, left a few small comments. Also why is wren added to this PR?

engine/renderer/private/gbuffers.cpp
MuskettaMan: code linting fix 😭

external/imgui_entt_entity_editor/imgui_entt_entity_editor.hpp
MuskettaMan: cause clang tidy sorts the includes

mmzala (approved):
Ah, so it was clang tidy that changed external files. I didn't think about that. LGTM now

PR #70, opened 2024-11-15
Feature/render module v2

MuskettaMan:
### Description

Second attempt at refactoring the renderer

Below is a list of things processed and todo, not everything has to be done in this PR but this is the state that it's currently at. Let me know what you guys think of the new architecture.

- [x] Moved renderer to its own module
- [x] This time all the features are still working
- [x] Removed reference passing and storing for VulkanBrain and replaced with shared_ptr
- [x] VulkanBrain replaced by VulkanContext
- [x] Removed resource management from the VulkanContext
- [x] Refactored architecture so a GraphicsContext is passed that is composed of a VulkanContext and GraphicsResources
- [x] Improved DrawStats to have an actual API
- [x] Created a class that handles the communication from vulkan and imgui, and exposes interface to receive imgui texture handles using our resource handles.
- [ ] Improve public API
  - [ ] BloomSettings should be split up into a settings struct and vulkan resources.
  - [ ] Split camera definitions into settings struct and vulkan resources
  - [ ] A more crititical look at the public API, ideally we dont have to pass any vulkan resources outside the module, and we can manage references using resource handles.
- [x] Add resource managemer for samplers
- [ ] Make resource manager handle fallbacks
- [x] Add reference counting and clean up for resources
  - [ ] Make sure resources arent deleted while the resource is still in a command buffer
- [ ] Remove shadow image from gbuffers
- [ ] Improve API for debug rendering
  - [ ] and look into removing the ImGui dependency
- [ ] Look into cross-dependency between renderer and engine modules
- [ ] Improve on when bindless descriptor set is updated
- [ ] Separate gltf model loading and graphics loading
- [ ] Implement abstractions for command buffers
  - [ ] Implicit draw call tracking
  - [ ] Resource handle tracking
- [ ] Make imgui seperate pass


engine/renderer/private/pipelines/ibl_pipeline.cpp
```diff
@@ -20,30 +32,26 @@ IBLPipeline::IBLPipeline(const VulkanBrain& brain, ResourceHandle<Image> environ
...
 {
-    const Image& irradianceMap = *_brain.GetImageResourceManager().Access(_irradianceMap);
-    const Image& prefilterMap = *_brain.GetImageResourceManager().Access(_prefilterMap);
+    const Image& irradianceMap = *_context->Resources()->ImageResourceManager().Access(_irradianceMap);
+    const Image& prefilterMap = *_context->Resources()->ImageResourceManager().Access(_prefilterMap);
 
-    util::BeginLabel(commandBuffer, "Irradiance pass", glm::vec3 { 17.0f, 138.0f, 178.0f } / 255.0f, _brain.dldi);
+    util::BeginLabel(commandBuffer, "Irradiance pass", glm::vec3 { 17.0f, 138.0f, 178.0f } / 255.0f, _context->VulkanContext()->Dldi());
 
     util::TransitionImageLayout(commandBuffer, irradianceMap.image, irradianceMap.format, vk::ImageLayout::eUndefined, vk::ImageLayout::eColorAttachmentOptimal, 6, 0, 1);
```
santiberna, line 56: having all of these essential commands inside of a util:: namespace is not as convenient as a dedicated CommandList class. It might be worth considering to create an abstractions that already has easier to use functions for common actions like: Setting pass name, setting barriers, binding descriptors.

engine/renderer/private/pipelines/shadow_pipeline.cpp
```diff
@@ -60,62 +69,65 @@ void ShadowPipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t cu
...
+
     uint32_t indirectCountOffset = scene.gpuScene->IndirectCountOffset();
-    commandBuffer.drawIndexedIndirectCountKHR(indirectDrawBuffer, 0, indirectCountBuffer, indirectCountOffset, scene.gpuScene->DrawCount(), sizeof(vk::DrawIndexedIndirectCommand), _brain.dldi);
-    _brain.drawStats.drawCalls++;
-    _brain.drawStats.indirectDrawCommands += scene.gpuScene->DrawCount();
-    _brain.drawStats.indexCount += scene.gpuScene->DrawCommandIndexCount();
+    commandBuffer.drawIndexedIndirectCountKHR(indirectDrawBuffer, 0, indirectCountBuffer, indirectCountOffset, scene.gpuScene->DrawCount(), sizeof(vk::DrawIndexedIndirectCommand), vkContext->Dldi());
 
-    commandBuffer.endRenderingKHR(_brain.dldi);
+    _context->GetDrawStats().IndirectDraw(scene.gpuScene->DrawCount(), scene.gpuScene->DrawCommandIndexCount());
```
santiberna, line 90: With a command list abstraction, we can track all of these draw statistic much more easily over every single command list and then sum them at the end of a frame for an accurate result (it is possible that someone forgets to update the draw stats after issueing a draw command)

engine/renderer/private/resource_management/mesh_resource_manager.cpp
```diff
@@ -0,0 +1,8 @@
+#include "resource_management/mesh_resource_manager.hpp"
+
+#include "gpu_resources.hpp"
+
+template <>
+std::weak_ptr<ResourceManager<Mesh>> ResourceHandle<Mesh>::manager = {};
+
+MeshResourceManager::MeshResourceManager() = default;
```
santiberna, line 8: The constructors to each resource manager should not be public. Instead, there should me a Make() static function that returns and assigns a new shared pointer instance as the global manager. This avoid the need to inherit from shared from this and forgetting to assign the global weak pointer to the manager. Personally, I am not fond of this method, but I realize that it is a good way to keep resource handle to 4 bytes

engine/renderer/private/vulkan_context.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+
+void VulkanContext::CreateInstance(const VulkanInitInfo& initInfo)
+{
+    CheckValidationLayerSupport();
+    if (ENABLE_VALIDATION_LAYERS && !CheckValidationLayerSupport())
+        throw std::runtime_error("Validation layers requested, but not supported!");
+
+    vk::ApplicationInfo appInfo {};
+    appInfo.pApplicationName = "";
```
santiberna, line 67: Could give a name to our engine rn...

engine/renderer/private/vulkan_context.cpp
```diff
@@ -0,0 +1,335 @@
...
+    commandPoolCreateInfo.flags = vk::CommandPoolCreateFlagBits::eResetCommandBuffer;
+    commandPoolCreateInfo.queueFamilyIndex = _queueFamilyIndices.graphicsFamily.value();
+
+    util::VK_ASSERT(_device.createCommandPool(&commandPoolCreateInfo, nullptr, &_commandPool), "Failed creating command pool!");
+}
+
+void VulkanContext::CreateDescriptorPool()
+{
+    std::vector<vk::DescriptorPoolSize> poolSizes = {
+        { vk::DescriptorType::eSampler, 1000 },
```
santiberna, line 285: maybe use powers of two instead of 1000?

engine/renderer/private/vulkan_context.cpp
```diff
@@ -0,0 +1,335 @@
...
+        { vk::DescriptorType::eStorageBuffer, 1000 },
+        { vk::DescriptorType::eUniformBufferDynamic, 1000 },
+        { vk::DescriptorType::eStorageBufferDynamic, 1000 },
+        { vk::DescriptorType::eInputAttachment, 1000 }
+    };
+
+    vk::DescriptorPoolCreateInfo createInfo {};
+    createInfo.poolSizeCount = poolSizes.size();
+    createInfo.pPoolSizes = poolSizes.data();
+    createInfo.maxSets = 200;
```
santiberna, line 301: 256 or use a constant defined in some header file?

engine/renderer/public/draw_stats.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <cstddef>
+
+class DrawStats
```
santiberna, line 5: Could be a part of a Command List abstraction to collect data implicitly on dispatch / draw calls

engine/renderer/public/frame_graph.hpp
```diff
@@ -38,31 +45,40 @@ using FrameGraphResourceHandle = uint32_t;
...
+        ResourceHandle<Buffer> handle = ResourceHandle<Buffer>::Null();
+        vk::PipelineStageFlags2 stageUsage;
     };
+
+    FrameGraphResourceInfo(const std::variant<std::monostate, StageBuffer, ResourceHandle<Image>>& resource)
+        : resource(resource)
+    {
+    }
+
+    std::variant<std::monostate, StageBuffer, ResourceHandle<Image>> resource;
```
santiberna, line 59: Yay for variants

santiberna (changes requested):
Great Work! I think this pull request is a good change of direction for keeping the renderer interface cohesive, especially the new resource management and handle classes. I have marked some specific improvement point in the comments on code. However, I took a couple notes that need an answer:
- The renderer module is not responsible for rendering the scene in its Tick method anymore, who is responsible for rendering the scene?
- There are some linter warnings and errors (like missing headers)

As well as some improvement points for the future (could be added as TODOs for later implementation):
- Resource Loading logic (stb image and fastgltf) can be moved to a dedicated Resources module (that could also include the definitions for resource handle and manager) Note: a module does not necessarily need a ModuleInterface, it could just define data structures and functions
- We need a renderer API that works with the ECS and deprecate Scene Description (HIGH PRIORITY)
- Could improve on Vulkan abstractions: We could have a command list class that keeps track of dispatched calls, removing the need for keeping track of DrawStats. We could also avoid passing the Graphics context to every vulkan class and settle only for passing the VkDevice for deallocation
- We can have a module that contains the Graphics Context and general vulkan abstractions and a Renderer Module that contains the specific pipelines and rendering data structs.

I think we should merge this after my above questions are answered and someone else probably takes a look through this...

engine/renderer/private/graphics_resources.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include "resource_management/buffer_resource_manager.hpp"
+#include "resource_management/image_resource_manager.hpp"
+#include "resource_management/material_resource_manager.hpp"
+#include "resource_management/mesh_resource_manager.hpp"
+
+GraphicsResources::GraphicsResources(const std::shared_ptr<VulkanContext>& vulkanContext)
+    : _vulkanContext(vulkanContext)
+{
+    _samplerResourceManager = std::make_shared<class SamplerResourceManager>(_vulkanContext);
```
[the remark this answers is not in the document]
MuskettaMan: Its needed cause of the duplicate naming of the method name. This way we dont need to prefix everything with Get

engine/source/editor.cpp
```diff
@@ -173,21 +146,20 @@ void Editor::Draw(PerformanceTracker& performanceTracker, BloomSettings& bloomSe
...
     ImGui::End();
-    //
 
     ImGui::Begin("Dump VMA stats");
 
     if (ImGui::Button("Dump json"))
     {
-        char* statsJson;
-        vmaBuildStatsString(_renderer.GetBrain().vmaAllocator, &statsJson, true);
+        // char* statsJson;
```
mmzala, line 154: Should the VMA stats dumping be commented?

mmzala (commented):
Aside from what Santiago already mentioned, I added 3 more suggestions

santiberna (approved):

mmzala (approved):

PR #71, opened 2024-11-20
Fix/bloom-lightning-pass

mmzala:
### Description

Brightness is now correctly loaded during the sky dome pass. Previously the brightness sampled from the lighting pass was discarded and overridden by the sky dome pass.

image

### Issues

_Provide a list of related issues that should be closed_

### Test criteria

1. Open the program
2. Look at the chessboard
3. Play around with the bloom strength
4. Notice that models also become brighter whenever the bloom strength increases


dumitrurares14 (approved):
Very nice 👍

PR #72, opened 2024-11-20
Feature/wren_interpreter

santiberna:
### Description

This PR introduces Wren as a third party dependency and creates the ScriptingContext and ScriptingModule classes.
Wren was just placed into the repository under external, since it does not support cmake (and it is a very small dependency), and I have created a CMakeLists.txt for it.

ScriptingModule is currently unused and there is no API, since this PR is just adding the ScriptingContext and its script interpreting features, there are NO gameplay differences in the engine.

The ScriptingContext will be the main interface for loading and interpreting wren scripts (called "modules"), as well as some QoL features for adding include paths, setting the output log for Wren scripting (std::cout by default) and some basic error handling. There are currently no C++ bindings.

One things to note:
- In order to make unit testing easier and more organized, you can now create a tests folder inside a module to add unit tests to it, which will be automatically added to the unit tests project. I have also ported the old tests to their respective modules

### Task

[Codecks Task](https://bubonic-brotherhood.codecks.io/card/19q-scripting-language)

### Test criteria

The Unit Tests should be tested on Windows and Linux, since otherwise this PR adds no functionality to the game (just the building blocks)


engine/scripting/public/scripting_module.hpp
```diff
@@ -0,0 +1,26 @@
...
+    void Shutdown(MAYBE_UNUSED Engine& engine) override {};
+
+public:
+    NON_COPYABLE(ScriptingModule);
+    NON_MOVABLE(ScriptingModule);
+
+    ~ScriptingModule() override = default;
+
+private:
+    std::unique_ptr<ScriptingContext> context {};
```
MuskettaMan, line 25: needs _

engine/scripting/tests/scripting_tests.cpp
```diff
@@ -0,0 +1,38 @@
+#include <gtest/gtest.h>
+#include <log.hpp>
+#include <scripting_context.hpp>
+#include <sstream>
+
+// Every test will initialize a wren virtual machine, better keep memory requirements low
+constexpr ScriptingContext::VMMemoryConfig MEMORY_CONFIG {
+    256ull * 4ull, 256ull, 50
```
MuskettaMan, line 8: 1024ull?

engine/utility/private/fileIO.cpp
MuskettaMan, line 1: file_io.cpp*

engine/utility/public/fileIO.hpp
MuskettaMan, line 1: file_io.hpp*

engine/utility/public/fileIO.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+constexpr auto TEXT_READ_FLAGS = std::ios::in;
+constexpr auto TEXT_WRITE_FLAGS = std::ios::out | std::ios::trunc;
+
+using FileTime = std::chrono::time_point<std::chrono::file_clock>;
+
+/// <summary>
+/// Open a file stream for reading. Specify 0 or std::ios::flags
+/// </summary>
+std::optional<std::ifstream> OpenReadStream(const std::string& path,
```
MuskettaMan, line 23: string_view preferred?

MuskettaMan (approved):
PR looks good, only comments are nitpicks and maybe some slight syntax improvements. Nice work! Cant wait to continue seeing this being developed

engine/utility/public/fileIO.hpp
santiberna: string view requires a little more error handling but I can try it

santiberna: I was able to use string_view for the external API, but I still require an internal conversion back to string for opening the streams (this is generally safer, since stringviews are not guaranteed to be null terminated)
```c++
std::optional<std::ifstream> fileIO::OpenReadStream(std::string_view path,
    std::ios::openmode flags)
{
    std::ifstream stream(std::string(path), flags);
    // ....
}
```
I don't think this change is worth it since it can lead to weird bugs and file paths need to be null terminated, which will either force checking logic on our end or creating an std::string out of the view

PR #73, opened 2024-11-21
Feature/directional light component

mmzala:
### Description

- Directional light is now a component in the ECS
- Added color functionality to the light
- Removed SceneDescription from GPUScene as it is no longer needed there

### Issues

[Task in Codecks](https://bubonic-brotherhood.codecks.io/milestones/run/40/card/15a-directional-light-component)

### Test criteria

1. Open the program
2. Inside the world inspector there should be an entity with the `Directional Light` name
3. Select it and play around with the values
4. The directional light should change accordingly


engine/renderer/private/gpu_scene.cpp
```diff
@@ -39,33 +41,59 @@ GPUScene::~GPUScene()
...
-
-    const glm::mat4 lightView = glm::lookAt(light.camera.position, light.camera.position - direction, glm::vec3(0, 1, 0));
-    glm::mat4 depthProjectionMatrix = glm::ortho(-light.camera.orthographicSize, light.camera.orthographicSize, -light.camera.orthographicSize, light.camera.orthographicSize, light.camera.nearPlane, light.camera.farPlane);
-    depthProjectionMatrix[1][1] *= -1;
-
-    DirectionalLightData& directionalLightData = sceneData.directionalLight;
-    directionalLightData.lightVP = depthProjectionMatrix * lightView;
-    directionalLightData.depthBiasMVP = DirectionalLight::BIAS_MATRIX * directionalLightData.lightVP;
-    directionalLightData.direction = glm::vec4(direction, light.shadowBias);
+    directionalLightView.each([this, &sceneData, &directionalLightIsSet](const auto directionalLightComponent, const auto transformComponent)
```
MuskettaMan, line 58: hmm can't we create a view and then use a for loop? I think it will look a little better instead of a lambda, and we can use a counter variable, since later we might support more then one directional light

MuskettaMan (approved):
Looks good! Only one comment about creating view, but nothing else that should change. For the time being have only one directional light is fine, I dont think we need more anyway

PR #75, opened 2024-11-25
Fix/breakpoints and tracy

santiberna:
Description
This pull request fixes three things:

Removes .run folder for CLion configurations (were not reliable since it kept being changed for no reason)
Tracy is disabled in Debug builds
FastGLTF is now using the main branch, which has the -Og flag propagation bug fixed
https://bubonic-brotherhood.codecks.io/card/18h-gcc-debugger-on-debug-config-still-optimizes-out-variables

(kinda this one, but also fixes slow startup and shutdown on unit tests)
https://bubonic-brotherhood.codecks.io/card/184-tracy-memory-crash

Test criteria
Preferably, this PR should be tested on Debug and Release and with Tracy, to make sure memory is profiled and things dont crash


dumitrurares14:
In release with tracy there is seems to be a bug where the engine crashes when you want to close it instead of actually closing it. You can replicate this by spawning some physics objects clearing them and then pressing ESC. it crashes at a worker thread

image

external/CMakeLists.txt
```diff
@@ -147,15 +153,15 @@ set(CPP_RTTI_ENABLED ON)
 set(ENABLE_ALL_WARNINGS OFF)
 
 # TODO: These options enable vectorization optimizations
-set(USE_SSE4_1 ON)
-set(USE_SSE4_2 ON)
-set(USE_AVX ON)
-set(USE_AVX2 ON)
+set(USE_SSE4_1 OFF)
+set(USE_SSE4_2 OFF)
```
dumitrurares14, line 157: we can keep this on but keep in mind that those were recommended from the jolt implementation from their example. If things break in jolt we should remember this place to come back later

external/CMakeLists.txt
```diff
@@ -171,20 +177,6 @@ target_compile_options(Jolt PRIVATE
         -Wno-error  # Do not treat warnings as errors
 )
 
-# GoogleTest
```
dumitrurares14, line 174: No more google test?

dumitrurares14 (changes requested):
Other than the crash the changes look solid, i did leave some questions around but maybe that's just me not understanding this fully. We should fix the crash before merging into main tho

external/CMakeLists.txt
santiberna: Wait I will check if I accidentally changed this

external/CMakeLists.txt
santiberna: Oh I am declaring Google test in the UnitTesting project instead of in the External one, should remove that comment

external/CMakeLists.txt
santiberna: My mistake, fixed it

santiberna: Yup, AVX512 was the only one disabled (seems like its a new flag and some processors might not support it)

dumitrurares14 (approved):
Very nice 🫡

PR #78, opened 2024-11-25
Physics Module now inherits from ModuleInterface

VengarvdWal:
The module is now properly initialized in the Engine.

https://bubonic-brotherhood.codecks.io/decks/7-engine/card/1cj-make-the-physics-system-a-module-as-i-have-never-used-the-cmake-modules

Test if the module still works the same as before. You can shoot raycasts with space, and the console outputs the hit info.


engine/source/engine.cpp
```diff
@@ -46,14 +46,14 @@ ModuleTickOrder OldEngine::Init(Engine& engine)
 
     auto& applicationModule = engine.GetModule<ApplicationModule>();
     auto& rendererModule = engine.GetModule<RendererModule>();
+    auto& physicsModule = engine.GetModule<PhysicsModule>();
 
     TransformHelpers::UnsubscribeToEvents(_ecs->registry);
     RelationshipHelpers::SubscribeToEvents(_ecs->registry);
     // modules
```
dumitrurares14, line 53: delete this comment now

dumitrurares14 (approved):
It looks good to me and it works as before so you get 100 points from me 🫡

PR #80, opened 2024-11-25
Feature/camera component

mmzala:
### Description

- Added camera component
- Removed old scene description
- Removed passing the camera structure to render passes via the constructor, as it is no longer needed

### Issues

[Codecks task](https://bubonic-brotherhood.codecks.io/milestones/run/40/card/1ck-camera-component)

### Test criteria

1. Open the application
2. Fly around the scene
3. The camera should work as previously

There is also one thing that I would like help with. In the `engine.cpp` file, I edited the camera movement, since the rotation is now saved as a quaternion. But clamping the pitch rotation no longer works, as when the yaw rotation is at 90 or -90 degrees, pitch snaps to 90 or -90 as well when clamping the pitch axis. Maybe I have overlooked something, so if anyone sees a mistake regarding it, that would be appreciated.


MuskettaMan (approved):
Looks good! Nice to see the ECS being fully in control now

PR #83, opened 2024-11-27
Feature/rendering particles

lyvanettinger:
### Description

BB-Prototype2024-11-2715-16-36-ezgif com-crop
- Added basic rendering for particles with a beautiful image
- Particle pipeline is in node in frame graph
- Particle system is integrated in renderer module
- Added slightly better random for directions of particles when spawning, but it's really not that great at all yet 🤡 there's a card already created to improve on this later on
- Added the `vk::DescriptorPoolCreateFlagBits::eUpdateAfterBind` flag during descriptorPool creation in vulkan context so descriptor set layouts can be created with this flag. This is needed so write descriptor sets can be updated after they're already bound while a commandbuffer is active each frame.
- Added deltaTime to RenderSceneDescription

### Issues

https://bubonic-brotherhood.codecks.io/card/19i-emitting-new-particles-sometimes-resets-particle-renders
https://bubonic-brotherhood.codecks.io/card/197-finish-up-basic-rendering

### Known concerns

- The particles seem semi-transparent which I suspect is because it's not doing anything regarding Bloom at all
- There's already a card for this on codecks, but at the moment all particle passes are in the same node in the frame graph rather than separated
- The particle system is integrated into the renderer module for now, but it's planned to be in its own module
- It's not stress tested yet on steam deck
- It's using drawIndexed at the moment as it was a little easier to implement than drawIndexedIndirect, but depending on profiling results this might be changed into drawIndexedIndirect

### Test criteria

- [ ] Press P to emit 5 particles at once
- [ ] Move the camera around to make sure the billboards are facing the camera at all times


engine/renderer/private/particles/particle_interface.cpp
```diff
@@ -30,6 +28,55 @@ ParticleInterface::ParticleInterface(const std::shared_ptr<ECS>& ecs)
...
+    emitter.velocity = glm::vec3(1.0f, 5.0f, 1.0f);
+    emitter.mass = 2.0f;
+    emitter.rotationVelocity = glm::vec3(1.0f);
+    emitter.maxLife = 5.0f;
+    emitter.materialIndex = LoadEmitterImage("assets/textures/nogameplay.png");
+    emitter.size = glm::vec2(resources->ImageResourceManager().Access(_emitterImages[0])->width, resources->ImageResourceManager().Access(_emitterImages[0])->height) / static_cast<float>(glm::max(resources->ImageResourceManager().Access(_emitterImages[0])->width, resources->ImageResourceManager().Access(_emitterImages[0])->height));
+    _emitterPresets.emplace_back(emitter);
+}
+
+uint32_t ParticleInterface::LoadEmitterImage(const char* imagePath)
```
santiberna, line 58: We should really make a global function for loading images... put a TODO

engine/renderer/private/particles/particle_pipeline.cpp
```diff
@@ -87,92 +123,159 @@ void ParticlePipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t
...
     util::BeginLabel(commandBuffer, "Particle rendering pass", glm::vec3 { 255.0f, 105.0f, 180.0f } / 255.0f, vkContext->Dldi());
+    commandBuffer.beginRenderingKHR(&renderingInfo, vkContext->Dldi());
+
+    commandBuffer.setViewport(0, 1, &_gBuffers.Viewport());
+    commandBuffer.setScissor(0, 1, &_gBuffers.Scissor());
 
     commandBuffer.bindPipeline(vk::PipelineBindPoint::eGraphics, _pipelines[static_cast<uint32_t>(ShaderStages::eRenderInstanced)]);
 
+    auto camera = scene.sceneDescription->camera;
+    glm::mat4 cameraRotation = mat4_cast(glm::quat(camera.eulerRotation));
```
santiberna, line 218: You forgot to include glm/quaternion.hpp

santiberna (commented):
Mini review, will test later

engine/renderer/private/particles/particle_interface.cpp
lyvanettinger: I added a card on Codecks for this https://bubonic-brotherhood.codecks.io/decks/7-engine/card/1e1-image-loader-utility-function

engine/renderer/private/particles/particle_interface.cpp
```diff
@@ -40,6 +87,7 @@ void ParticleInterface::SpawnEmitter(EmitterPreset emitterPreset, uint32_t times
         {
             emitterComponent.timesToEmit = timesToEmit;
             emitterComponent.emitter = _emitterPresets[static_cast<int>(emitterPreset)];
+            spdlog::info("Spawned emitter!");
```
MuskettaMan, line 90: Cna probably remove this, so we dont clutter the logs

engine/renderer/private/particles/particle_pipeline.cpp
```diff
@@ -87,92 +125,159 @@ void ParticlePipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t
...
     util::EndLabel(commandBuffer, vkContext->Dldi());
+}
+
+void ParticlePipeline::RecordRenderIndexed(vk::CommandBuffer commandBuffer, uint32_t currentFrame, const RenderSceneDescription& scene)
+{
+    auto vkContext { _context->VulkanContext() };
+    auto resources { _context->Resources() };
+    auto culledIndicesBuffer = resources->BufferResourceManager().Access(_culledInstancesBuffer);
+
+    // make sure the compute is done before the host reads from it
```
MuskettaMan, line 172: Wait do we read data back from the device to the host? That sounds slow...

engine/renderer/private/particles/particle_pipeline.cpp
```diff
@@ -87,92 +125,159 @@ void ParticlePipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t
...
+    commandBuffer.beginRenderingKHR(&renderingInfo, vkContext->Dldi());
+
+    commandBuffer.setViewport(0, 1, &_gBuffers.Viewport());
+    commandBuffer.setScissor(0, 1, &_gBuffers.Scissor());
 
     commandBuffer.bindPipeline(vk::PipelineBindPoint::eGraphics, _pipelines[static_cast<uint32_t>(ShaderStages::eRenderInstanced)]);
 
+    auto camera = scene.sceneDescription->camera;
+    glm::mat4 cameraRotation = mat4_cast(glm::quat(camera.eulerRotation));
+    glm::mat4 cameraTranslation = translate(glm::mat4 { 1.0f }, camera.position);
```
MuskettaMan: is there a using statement for glm somewhere? Make sure to use the full glm::

engine/renderer/private/particles/particle_pipeline.cpp
```diff
@@ -87,92 +125,159 @@ void ParticlePipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t
...
+    auto cameraView = glm::inverse(cameraTranslation * cameraRotation);
+    _renderPushConstant.cameraRight = glm::vec3(cameraView[0][0], cameraView[1][0], cameraView[2][0]);
+    _renderPushConstant.cameraUp = glm::vec3(cameraView[0][1], cameraView[1][1], cameraView[2][1]);
+    commandBuffer.pushConstants(_pipelineLayouts[static_cast<uint32_t>(ShaderStages::eRenderInstanced)], vk::ShaderStageFlagBits::eVertex, 0, sizeof(RenderPushConstant), &_renderPushConstant);
```
MuskettaMan: use the modern overload of this function, something like
`commandBuffer.pushConstants<RenderPushConstant>(_pipelineLayouts[static_cast<uint32_t>(ShaderStages::eRenderInstanced)], vk::ShaderStageFlagBits::eVertex, 0, { _renderPushConstant });`

engine/renderer/private/particles/particle_pipeline.cpp
```diff
@@ -87,92 +125,159 @@ void ParticlePipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t
...
-
-        commandBuffer.pipelineBarrier(vk::PipelineStageFlagBits::eTransfer, vk::PipelineStageFlagBits::eComputeShader, vk::DependencyFlags { 0 }, {}, barrier, {});
+        util::CopyBuffer(commandBuffer, _stagingBuffer, resources->BufferResourceManager().Access(_emittersBuffer)->buffer, bufferSize);
     }
 }
 
+void ParticlePipeline::UpdateAliveLists()
+{
+    std::swap(_particlesBuffers[static_cast<uint32_t>(ParticleBufferUsage::eAliveNew)], _particlesBuffers[static_cast<uint32_t>(ParticleBufferUsage::eAliveCurrent)]);
+    UpdateParticleBuffersDescriptorSets();
```
MuskettaMan, line 266: I would still make a task or improvement for making this data use a bindless model. That would remove all this descriptor bloat and make it simpler for swapping, since you can pass the indices as a push constant

engine/renderer/private/particles/particle_pipeline.cpp
```diff
@@ -291,7 +396,12 @@ void ParticlePipeline::CreatePipelines()
         pipelineLayoutCreateInfo.setLayoutCount = descriptorLayouts.size();
         pipelineLayoutCreateInfo.pSetLayouts = descriptorLayouts.data();
 
-        pipelineLayoutCreateInfo.pushConstantRangeCount = 0;
+        vk::PushConstantRange pcRange {};
+        pcRange.stageFlags = vk::ShaderStageFlagBits::eVertex;
```
MuskettaMan: You should use the PipelineBuilder class for generating this graphics pipeline, it makes it much shorter and removes a lot of bloat. It also automatically infers vertex structure from the spv binaries. The builder will also constructor descriptor layouts for you, but this can be annoying, so after you make your descriptor set layout you can cache it using a static function on the pipeline builder.

engine/renderer/private/particles/particle_pipeline.cpp
```diff
@@ -695,44 +814,64 @@ void ParticlePipeline::CreateBuffers()
...
+        _culledInstancesBuffer = resources->BufferResourceManager().Create(creation);
+        cmdBuffer.CopyIntoLocalBuffer(culledInstances, 0, resources->BufferResourceManager().Access(_culledInstancesBuffer)->buffer);
     }
 
-    { // Culled Instance SSB
-        std::vector<uint32_t> culledInstance(MAX_PARTICLES + 1);
-        vk::DeviceSize culledInstanceBufferSize = sizeof(uint32_t) * (MAX_PARTICLES + 1);
+    { // Billboard vertex buffer
+        std::vector<Vertex> billboardPositions = {
+            // TODO: tangents..?
```
MuskettaMan, line 832: only required for normal maps

MuskettaMan (changes requested):

engine/renderer/private/particles/particle_pipeline.cpp
lyvanettinger: Yeah at the moment we do for the drawIndexed to know how many instances we need. Assuming we'll transition to drawIndirect (which I would personally want to at some point) this readback should be eliminated with that

MuskettaMan (approved):

PR #85, opened 2024-11-28
Feature/wren bindings

santiberna:
### Description

This pull requests adds an API for the engine to expose C++/Wren bindings. In order to do this in a type safe way and without reinventing the wheel, I have used instead the wrenbind17 library, that exposes templates for exposing C++ types to wren. This pull request also adds the foundation for a main wren entry point and update function, but it is still does not support hot reloading and such.

There are still no bindings for any engine modules, except the time module which was a good testing target. Every module that wishes to expose bindings must use the wrenbind api to do so:
```c++

// Example binding for vec3

auto& v3 = module.klass<glm::vec3>("Vec3");

// wrenbind only supports one constructor
// we can use static functions for the others
auto& foreignAPI = vm.module("Engine");
auto& v3 = foreignAPI.klass<glm::vec3>("Vec3");

v3.ctor<float, float, float>();
v3.funcStatic<bind::Vec3Identity>("Identity");
v3.var<&glm::vec3::x>("x");
v3.var<&glm::vec3::y>("y");
v3.var<&glm::vec3::z>("z");
v3.funcExt<bind::Vec3ToString>("ToString");

// Example binding for TimeModule

auto& foreignApi = e.GetModule<ScriptingModule>().GetForeignAPI();
auto& wrenClass = foreignApi.klass<TimeModule>("TimeModule");
wrenClass.funcExt<bind::TimeModuleGetDeltatime>("GetDeltatime");

```

All bindings should be done in the respective module's Init function (since access to the ScriptingModule is necessary)
In order to make a module accessible through the engine, you must also add it in through the `BindModule<T>` functions of the ScriptingModule:

```c++
auto& scripting = instance.GetModule<ScriptingModule>();

// Add modules here to expose them in scripting
{
    auto& engineAPI = scripting.StartEngineBind();
    scripting.BindModule<TimeModule>(engineAPI, "Time");
    scripting.EndEngineBind(instance);
}
``` 

This must happen after the engine is initialized in `main`, as the order in which this happens is important.

There is a sample Game.wren script that simply prints the amount of frames per second

### Issues

https://bubonic-brotherhood.codecks.io/card/1az-wren-foreign-binding-api

### Test criteria

Run the engine and check if frames per second are printed to the screen.
I also believe the code here is a little messy and I can probably better organize some sections of the code (reviews are welcome to show potential improvements)


engine/core/public/time_module.hpp
MuskettaMan, line 1: check for snake case

engine/scripting/private/utility/foreign_bindings.cpp
```diff
@@ -0,0 +1,48 @@
+#include "utility/foreign_bindings.hpp"
+#include "log.hpp"
+#include "utility/script_file_loading.hpp"
+
+ForeignBindings::ForeignBindings(const std::string& engineModule)
+{
+    _engineModule = ScriptFileLoading::CanonicalizePath(engineModule);
```
MuskettaMan, line 7: can be done in initializer

engine/scripting/public/scripting_module.hpp
```diff
@@ -1,26 +1,64 @@
...
+        WrenEngine::BindModule<T>(e, name);
+    }
+
+    void EndEngineBind(Engine& e) const
+    {
+        WrenEngine::BindEngineEnd(GetForeignAPI(), _context->GetVM(), e);
+    }
+
 private:
+    ForeignBindings _foreign_bindings { "EngineAPI" };
```
MuskettaMan, line 60: snake

game/engine_api.wren
MuskettaMan, line 1: If this is generated, should it be gitignored?

MuskettaMan (changes requested):
Minor comments

game/engine_api.wren
santiberna: Yeah will add it

engine/scripting/public/scripting_module.hpp
santiberna: God if I didn't have you

engine/scripting/private/utility/foreign_bindings.cpp
santiberna: For sure... but I think I will rework this class a bit since its a bit convoluted

MuskettaMan (approved):
Looks good, discussed internally and approach looks managable and extendable

PR #88, opened 2024-11-28
Particle Module

lyvanettinger:
### Description

- Moved files related to particles to new particle module
- Added ParticleModule module

### Issues

https://bubonic-brotherhood.codecks.io/card/191-make-particle-system-a-module

### Test criteria

- [ ] Particles can still be spawned and seen after pressing P


engine/particles/private/particle_util.cpp
MuskettaMan, line 1: can be removed, if empty right?

VengarvdWal (commented):
So if I get it right, this is not a standalone module but a "sub" module of the renderer?

Because the particles do not have a "module" or a class that inherits from ModuleInterface. I don't know if this intented, because why couldn't we just get the particle module from the engine in the renderer?

VengarvdWal (changes requested):
I would solve some of the linter errors, as these can still give some nasty bugs later on. I am still not sure about the pipeline.hpp being moved back to the particle folder instead of the pipeline folder. 

Also the Module structure is a little bit odd, because it does not follow the module principle with overriding ModuleInterface and adding it as a module to the engine main.cpp. But then again I am not sure what was discussed before about making it part of the Renderer Module.

VengarvdWal (approved):
Looks good

PR #89, opened 2024-11-28
Feature/load and play sounds

LeoHellmig:
### Description

Loading and playing sounds in the audio module.
We can now:
-  load .mp3 files by using LoadSFX and play them using PlaySFX
-  load fmod studio .bank files using LoadBank
-  play fmod events using StartEvent

### Issues

https://bubonic-brotherhood.codecks.io/card/1af-audio-module 

### Test criteria
- [x] Test the sound banks loading by taking a look in the terminal and watch out for any errors
- [x] Press or hold 'o' to play a shooting sound
- [ ] Could be tested on Steamdeck, but it ran without issues on WSL for me


engine/audio/private/audio_module.cpp
```diff
@@ -1,68 +1,178 @@
...
-    if (result != FMOD_OK)
-    {
-        bblog::error("FMOD Error: {0}", FMOD_ErrorString(result));
-        return tickOrder;
-    }
+        CHECKRESULT(FMOD_Studio_System_GetCoreSystem(_studioSystem, &_coreSystem));
 
-    result = FMOD_Studio_System_GetCoreSystem(_studioSystem, &_coreSystem);
-    if (result != FMOD_OK)
+        CHECKRESULT(FMOD_System_GetMasterChannelGroup(_coreSystem, &_masterGroup));
```
lyvanettinger, line 26: The naming is quite general. Something like FMOD_ASSERT might be better

lyvanettinger (commented):
A small comment on naming, other than that looks good so far. Will test the branch still

lyvanettinger (changes requested):
image

Is there any way to clean up the log a little? It appears this block of text is logged for each sound loaded, and when it's unloaded it gets logged again. These clutter up our logs a lot making it harder to tell what's necessary to see e.g. (validation) error messages. If you still need this for debugging sometimes is it possible to make the logging optional?

assets/sounds/Master.bank
santiberna, line 1: add assets/**/*.bank to the gitattributes lfs

engine/audio/private/fmod_debug.cpp
```diff
@@ -4,7 +4,14 @@
 #include "fmod_errors.h"
 #include "log.hpp"
 #include <iostream>
-#include <string>
+void CHECKRESULT_fn(FMOD_RESULT result, MAYBE_UNUSED const char* file, int line)
```
santiberna, line 7: Forgot to include common.hpp for the MAYBE_UNUSED macro. Also should rename to FMODCheckResult since other libraries can also have handling for their result types

engine/audio/private/fmod_debug.cpp
```diff
@@ -4,7 +4,14 @@
...
+{
+    if (result != FMOD_OK)
+    {
+        bblog::error("FMOD ERROR: audio_module.cpp [Line {0} ] {1} - {2}", line, static_cast<int>(result), FMOD_ErrorString(result));
+        throw FMOD_ErrorString(result);
```
santiberna, line 12: Might be better to throw std::runtime_error(FMOD_ErrorString(result)). This allows us to handle exceptions like this with try catch and probably prints better debug messages

engine/audio/private/fmod_debug.hpp
```diff
@@ -1,3 +1,6 @@
 #pragma once
 
+void CHECKRESULT_fn(FMOD_RESULT result, MAYBE_UNUSED const char* file, int line);
```
santiberna, line 3: You should include the header that has FMOD_RESULT

santiberna (changes requested):
Couple notes, but otherwise seems fine

lyvanettinger (approved):
Looks good, nice job!

PR #92, opened 2024-12-02
Features/colliders

dumitrurares14:
### Description

Added physics colliders to meshes. A quick auxiliary component was created to the ecs to the entities that have a mesh component to save the mesh bounds easily.

Here is a video of the physics colliders spawned into the chess scene where we have a somewhat complex hierarchy. I used the ray shooting functions to combine everything into one just to prove everything works nicely.

https://github.com/user-attachments/assets/ba8db054-4011-4077-8369-d9ceeb363ec2

### Issues

https://bubonic-brotherhood.codecks.io/card/16r-physics-system-complex-colliders

### Test criteria

- [ ] I'd appreciate any feedback on the way I approached this with that auxiliary component.
- [ ] Open it and see if it runs and performs nicely on your machine too. Try more scenes if you have any besides the chess one and the monkeys one that luuk made.


engine/source/scene_loader.cpp
```diff
@@ -12,34 +13,46 @@
...
+    TransformHelpers::SetLocalTransform(ecs.registry, entity, currentNode.transform);
+
     if (currentNode.meshIndex.has_value())
     {
         ecs.registry.emplace<StaticMeshComponent>(entity).mesh = model.meshes.at(currentNode.meshIndex.value());
+
+        TempPhysicsData tempData;
+        tempData.boundingBox = cpuModel.meshes.at(currentNode.meshIndex.value()).GetMeshBounds();
+
+        ecs.registry.emplace<TempPhysicsData>(entity, tempData);
```
santiberna, line 42: Regarding using two components for Bounding boxes and StaticMeshComponent, why do we not simply store the data along with a Rigidbody component or StaticMeshComponent

santiberna (changes requested):
I think you should revisit the use of TempPhysicsData, since I believe you can just put that data inside of the Static Mesh component. Otherwise, looks good with a couple options.

engine/source/scene_loader.cpp
dumitrurares14: I decided to store it in the static mesh component in the end, Rigid body component is not a good choice since we don't want to create rigid body components in the model loader.

engine/source/engine.cpp
```diff
@@ -75,17 +76,18 @@ ModuleTickOrder OldEngine::Init(Engine& engine)
...
 
+    /*
     TransformHelpers::SetLocalRotation(_ecs->registry, entities[0], glm::angleAxis(glm::radians(45.0f), glm::vec3 { 0.0f, 1.0f, 0.0f }));
     TransformHelpers::SetLocalPosition(_ecs->registry, entities[0], glm::vec3 { 10.0f, 0.0f, 10.f });
 
     TransformHelpers::SetLocalScale(_ecs->registry, entities[1], glm::vec3 { 4.0f });
     TransformHelpers::SetLocalPosition(_ecs->registry, entities[1], glm::vec3 { 106.0f, 14.0f, 145.0f });
 
-    TransformHelpers::SetLocalPosition(_ecs->registry, entities[2], glm::vec3 { 20.0f, 0.0f, 20.0f });
+    TransformHelpers::SetLocalPosition(_ecs->registry, entities[2], glm::vec3 { 20.0f, 0.0f, 20.0f });*/
```
santiberna, line 90: Remove or uncomment

santiberna (approved):
Check the commented sections and remove if needed. Otherwise, merge it!

PR #93, opened 2024-12-02
[FIXED] Exit early if no materials present to override

MuskettaMan:
### Description

Fixes vulkan validation error with no materials being present

### Test criteria

Make sure no vulkan validation errors appear in the settings.

- [x] Is tested on steam deck


santiberna (approved):
Looks good, but you could also fix the linter warnings as well

PR #94, opened 2024-12-03
Feature/sdl controller input

mmzala:
### Description

Adds controller support using SDL.

- Button inputs (up/down/held)
- Axis and triggers inputs (from -1 to 1 mapping)
- Deadzone handling

This PR adds the functionality directly to the input manager. The next PR will include Steam Input API and some sort of abstraction for the controller input based on if Steam Input is available or not. If Steam Input is available, we will use that for controllers, otherwise SDL will be used.

### Issues

[Codecks card](https://bubonic-brotherhood.codecks.io/milestones/run/41/card/19f-controller-input-sdl3)

### Test criteria

1. Make sure you have a controller connected (steam deck should also work)
2. Open the program
3. Fly around with the camera using the axis inputs


MuskettaMan (approved):
Looks good, nice interface, I have some minor comments, but it should be good besides it

PR #96, opened 2024-12-04
Feature/ecs module

santiberna:
### Description

Makes the ECS into a module, meaning that Renderer and Particle don't need to depend on OldEngine for the ECS. I still think this is very messy since I was forced to add a ECSModule& to a lot of classes, where maybe it made more sense to pass the registry to the functions that needed it.

Also moved most the components in old engine to Renderer or ECS module folders

### Issues

[Codecks Issue](https://bubonic-brotherhood.codecks.io/card/1ez-ecs-registry-module)

### Test criteria

The engine should behave the exact same as before. Check if game runs correctly and physics and particles are working as expected


engine/ecs/private/ecs_module.cpp
```diff
@@ -0,0 +1,55 @@
...
+void ECSModule::Shutdown(MAYBE_UNUSED Engine& engine)
+{
+    TransformHelpers::UnsubscribeToEvents(registry);
+    RelationshipHelpers::UnsubscribeToEvents(registry);
+}
+
+void ECSModule::Tick(Engine& engine)
+{
+    // auto dt = engine.GetModule<TimeModule>().GetDeltatime().count();
+    float dt = 16.66f;
```
lyvanettinger, line 21: Forgot to uncomment getting deltaTime from TimeModule?

engine/source/engine.cpp
```diff
@@ -306,11 +304,6 @@ void OldEngine::Tick(Engine& engine)
         audioModule.StopEvent(eventId);
     }
 
-    _ecs->UpdateSystems(deltaTimeMS);
-    _ecs->GetSystem<PhysicsSystem>()->CleanUp();
```
lyvanettinger, line 310: Might've forgotten to put the physicssystem cleanup in the render module's tick

lyvanettinger (changes requested):
Some comments on stuff I noticed on first read through. Will still test the branch

engine/ecs/private/ecs_module.cpp
santiberna: Yeah did, good catch!

engine/source/engine.cpp
santiberna: Yeah I have just noticed... do you think there is a way to not explicitly place it in the ECS module? Id rather not add a dependency on the OldEngine for this cleanup function

lyvanettinger: I'd probably suggest moving the physics system (same for the rigidbody component) away from OldEngine since it doesn't really make since being in that place anyway now that the ECS has moved. Judging from particles and renderer modules which have their components in their own modules, I'd suggest moving the physics system and rigidbody component to the physics module.

lyvanettinger: This would mean that the ECS will have a dependency on the physics module which I hope isn't too big of an issue

santiberna: Oh thats a nice idea... will try it.

PR #97, opened 2024-12-05
Feature/math bindings

santiberna:
### Description

Mini pull request, exposing glm types in the wren scripting side. Also I tried adding the essential math types and operations, but if there are any that I missed please comment

### Issues

No issues but this will be useful for creating some gameplay

### Test criteria

Not a lot of code, might be able to just check if some of the operations work, but I trust wrenbind on this


MuskettaMan (approved):
Tested in engine and looks good!

PR #98, opened 2024-12-05
Feature/point lights

VengarvdWal:
### Description

Point Light implementation. You can now create point lights in the game, either during runtime with the "inspector" just add a component. Or creating entities and adding the component there in engine.cpp, or other place. The buffer is a storage buffer that can currently hold 10240 max point lights. 

pointlights_in_engine
pointlights_in_engine_v2

I need point lights for the Clustered Rendering, I am also working on.

### Issues
https://bubonic-brotherhood.codecks.io/card/1fq-add-point-light-component-etc

### Test criteria
- [ ] Lighting still works as intended
- [ ] You can add lights with the inspector, or creating them in code. 
- [ ] Colors can be changed via the inspector at runtime.


engine/renderer/private/gpu_scene.cpp
```diff
@@ -149,6 +161,29 @@ void GPUScene::UpdateDirectionalLightData(SceneData& scene, uint32_t frameIndex)
...
+    for (const auto& [entity, pointLightComponent, transformComponent] : pointLightView.each())
+    {
+        if (pointLightCount >= MAX_POINT_LIGHTS)
+        {
+            bblog::warn("Reached the limit of point lights available");
+            break;
+        }
+
+        PointLightData& pointLightData = pointLightArray.lights[pointLightCount];
+        pointLightData.position = glm::vec4(TransformHelpers::GetLocalPosition(transformComponent), 1.0f);
```
MuskettaMan, line 178: Local position? Why not world position, since this would break hierarchy, right? World position is free anyway since we have the world matrix cached

engine/renderer/public/gpu_scene.hpp
```diff
@@ -32,6 +32,7 @@ struct RenderSceneDescription
 };
 
 constexpr uint32_t MAX_INSTANCES = 2048;
+constexpr uint32_t MAX_POINT_LIGHTS = 10240;
```
MuskettaMan, line 35: why not a power of two? its more satisfying and plays nicer with memory sizes

engine/source/editor.cpp
```diff
@@ -28,7 +28,9 @@
 #include <imgui/misc/cpp/imgui_stdlib.h>
 
 // TODO: Editor shouldnt depend on this.
```
MuskettaMan, line 30: can remove todo. I added it before, but it isnt really something we should concern us with anymore

shaders/lighting.frag
```diff
@@ -90,19 +95,47 @@ void main()
...
+        PointLight light = pointLights.lights[i];
+        vec3 lightPos = light.position.xyz;
+        vec3 L = normalize(lightPos - position);
+        float distance = length(lightPos - position);
+        float attenuation = clamp(1.0 - (distance / light.range), 0.0, 1.0);
+
+        vec3 H = normalize(V + L);
+        vec3 F = FresnelSchlick(max(dot(H, V), 0.0), F0);
+
+        float NDF = DistributionGGX(N, H, roughness);
```
MuskettaMan, line 110: from what i can tell a lot of this code is duplicated from the directional light calculation. I think we can generalize this to a function, right?

shaders/scene.glsl
```diff
@@ -1,3 +1,5 @@
+#define MAX_POINT_LIGHTS 10240
```
MuskettaMan, line 1: make sure this matches with the new power of two point light amount

MuskettaMan (commented):
Looks good! Few comments to clean it up and make it ready to merge

engine/renderer/private/gpu_scene.cpp
VengarvdWal: How would I do this? Because the returns a mat4, are there any examples that use this?

engine/ecs/private/components/point_light_component.cpp
```diff
@@ -0,0 +1,18 @@
...
+namespace EnttEditor
+{
+template <>
+void ComponentEditorWidget<PointLightComponent>(entt::registry& reg, entt::registry::entity_type e)
+{
+    auto& t = reg.get<PointLightComponent>(e);
+    t.Inspect();
+}
+}
+void PointLightComponent::Inspect()
```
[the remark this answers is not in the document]
VengarvdWal: So do I just remove it then? Because I thought being able to edit the pointlight data during runtime would be good. Or do you want to move this functionality somewhere else? I am not familiar with the Editor things yet

shaders/lighting.frag
VengarvdWal: I refactored the lighting shader. Much cleaner now in my opinion. But let me know what you think of it.

MuskettaMan (approved):
Looks good!

PR #99, opened 2024-12-09
Feature/steam input

mmzala:
### Description

- Added Steam Input
- Added Input Actions for both Steam and SDL devices
- Reworked how Input Devices are handled
- Added Steam Input configuration files (which should be officially uploaded to Steamworks later on)
- Added Actions based on Pawel's controller scheme
- Converted camera movement to use actions

Includes both Steam Input and Input Actions, which should have been split up.... But without Input Actions, Steam Input wouldn't be implemented well, as it only works with Input Actions. So it's quite a big PR, here is an overview how the structure looks like:

InputHnadling drawio
Digital actions -> Actions involving button inputs
Analog actions -> Action involving axis inputs

### Issues

[Steam input codecks card](https://bubonic-brotherhood.codecks.io/decks/30-controls/card/18q-steam-input)
[Input Actions codecks card](https://bubonic-brotherhood.codecks.io/decks/30-controls/card/1fr-input-actions)

### Test criteria

With Steam (Steam Input Action will be used):
1. Make sure Steam is open in the background
2. Start our program with a controller connected (this is the tricky part as I don't know if all controllers will work as I only could set up the generic controller, each controller type can have their own configuration file file action inputs for Steam)
3. Fly around the scene with the axis inputs, it should work as previously

Without Steam (SDL Input Action will be used):
1. Start our program with a controller connected
2. Fly around the scene with the axis inputs, it should work as previously


engine/application/public/input/action_manager.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+using DigitalInputActionList = std::vector<DigitalInputAction>;
+using AnalogInputActionList = std::vector<AnalogInputAction>;
+
+// Action for button inputs.
+struct DigitalAction
+{
+    std::string name {};
+    DigitalActionType type {};
+    DigitalInputActionList inputs {};
```
santiberna, line 31: input bindings is probably more appropriate for a name here, instead of input action (DigitalInputAction -> DigitalInputBinding, AnalogInputAction -> AnalogInputBinding).

engine/source/engine.cpp
```diff
@@ -128,16 +129,80 @@ ModuleTickOrder OldEngine::Init(Engine& engine)
     audioModule.LoadBank(stringBank);
     audioModule.LoadBank(bi);
 
+    // Setting up game input actions
+    GameActions gameActions {};
```
santiberna, line 133: Is this entire section of setup also necessary for when working with steam input, or do the input scheme files already handle this?

engine/source/engine.cpp
```diff
@@ -128,16 +129,80 @@ ModuleTickOrder OldEngine::Init(Engine& engine)
...
+    reloadAction.type = DigitalActionType::ePressed;
+    reloadAction.inputs.emplace_back(GamepadButton::eWEST);
+
+    DigitalAction& jumpAction = actionSet.digitalActions.emplace_back();
+    jumpAction.name = "Jump";
+    jumpAction.type = DigitalActionType::ePressed;
+    jumpAction.inputs.emplace_back(GamepadButton::eSOUTH);
+
+    DigitalAction& meeleeAction = actionSet.digitalActions.emplace_back();
+    meeleeAction.name = "Meelee";
```
santiberna, line 190: melee*

santiberna (changes requested):
Mini requests and questions on the PR. Overall, it is a pretty nice implementation that handles both platforms. Good work on figuring out the steam input situation, since I believe it is hard to make SDL work cleanly with.
 
It might also be better to place the controller setup in its own header file, since then it is in a place where everyone can consult and modify

Personal opinion incoming: I do think you can probably use less `using` statements, especially when it's just over an ``std::vector<T>`` It makes the code a little harder to read since it looks like another struct

engine/application/CMakeLists.txt
```diff
@@ -3,7 +3,9 @@ module_default_init(Application)
 
 target_link_libraries(Application
         PUBLIC Core
+        PUBLIC Steam
```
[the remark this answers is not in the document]
mmzala: There are that's why I left it there :)

engine/application/public/input/action_manager.hpp
mmzala: You're right, naming is hard. I'll change it

engine/application/CMakeLists.txt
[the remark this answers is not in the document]
santiberna: Oh there were errors? Thats alright then

engine/source/engine.cpp
mmzala: All of the actions and action sets with the correct names still needs to be set up. Since those names are used to check for Steam Input actions that exist in the Steam Input files. Then Steam checks the inputs from the controller based on the files given. It is an extra layer of indirection needed, but it adds the possibility for the user to change what inputs map to what actions via the Steam Overlay.

PR #100, opened 2024-12-09
Feature/ecs bindings

santiberna:
### Description

This Pull request adds a new Bindings module, as well as bindings for working with the ECS. Due to some limitations around exposing things for wren, all code that deals with exposing variables, components and Engine modules should be here.

For the ECS, I have exposed functions to create entities and get an entity by name, as well as destroying entities.
Each entity is able to get their respective components (which is different from how it works in C++ but is a little easier to work with). 

### Issues

[Codecks Task](https://bubonic-brotherhood.codecks.io/card/1es-wren-bindings-for-ecs)

### Test criteria

The player should spawn slightly above ground now (the main script is modifying the initial player position). There should be not exceptions or logs about Wren compilation.

What follows is the current main wren script for showing the API:

```wren
import "engine_api.wren" for Engine, TimeModule, ECS, Entity, Vec3, TransformComponent

class Main {

    static Start(engine) {
        __counter = 0
        __frameTimer = 0
        __player = engine.GetECS().GetEntityByName("Camera")

        if (__player) {
            System.print("Player is online!")

            var t = __player.GetTransformComponent()
            t.translation = Vec3.new(0.0, 5.0, 0.0)
        }
    }

    static Update(engine, dt) {
        __counter = __counter + 1

        var deltatime = engine.GetTime().GetDeltatime()
        __frameTimer = __frameTimer + dt

        if (__frameTimer > 1000.0) {
            System.print("%(__counter) Frames per second")
            __frameTimer = __frameTimer - 1000.0
            __counter = 0
        }
    }
}
```


LarsJanssen21 (approved):
Does exactly what is described!
Using different values in the wren file's translation component reflects back on game!
One error in the output log for steamworks but that doesn't pertain to this PR

PR #102, opened 2024-12-10
Particle module interface

lyvanettinger:
### Description

For visuals, check out the show and tell channel for Compute Particles on discord!

- Removed old particle interface
- Implemented new design
- Added two tests:
   - Press space to shoot rays and when it hits, it spawns an emitOnce emitterComponent
   - Press 0 to spawn a physics entity with an emitterComponent attached emitting particles over time

### Issues

https://bubonic-brotherhood.codecks.io/card/1e8-implement-new-shell-of-interface

### Test criteria

- [x] Press 0 to spawn physics entity with emitter component (be aware that this quickly falls below the map so for best visuals, go underneath the map and then spawn it)
- [x] Press space near a collider to shoot a ray at it and on hit, emit particles once
- [x] Take a look at the particle module's Tick and SpawnEmitter functions in code; could it be done better? Could it break in any way?


engine/particles/private/particle_module.cpp
```diff
@@ -1,17 +1,177 @@
...
+
+    return ModuleTickOrder::ePostRender;
+}
+
+void ParticleModule::Tick(MAYBE_UNUSED Engine& engine)
+{
+    const auto emitterView = _ecs->GetRegistry().view<EmitterComponent>();
+    for (const auto entity : emitterView)
+    {
+        if (_ecs->GetRegistry().all_of<RigidbodyComponent>(entity))
```
MuskettaMan, line 33: maybe i dont understand `all_of` but why not add Rigidbodycomponent to the view?

engine/particles/private/particle_module.cpp
```diff
@@ -1,17 +1,177 @@
...
+        {
+            _ecs->GetRegistry().remove<EmitterComponent>(entity);
+            _ecs->GetRegistry().remove<ActiveEmitterTag>(entity);
+            continue;
+        }
+        if (_ecs->GetRegistry().all_of<RigidbodyComponent>(entity))
+        {
+            const auto& rb = _ecs->GetRegistry().get<RigidbodyComponent>(entity);
+
+            // TODO: this is temporary to avoid when an entity doesn't have a WorldMatrixComponent. Would there be a way to do this differently?
```
MuskettaMan, line 59: Doesn't every entity in the hierarchy guaranteed have a world matrix component?

engine/particles/private/particle_module.cpp
```diff
@@ -1,17 +1,177 @@
...
+                emitter.emitter.velocity = -glm::vec3(rbVelocity.GetX(), rbVelocity.GetY(), rbVelocity.GetZ());
+            }
+        }
+
+        // update timers
+        if (emitter.currentEmitDelay < 0.0f)
+        {
+            emitter.currentEmitDelay = emitter.maxEmitDelay;
+        }
+        emitter.currentEmitDelay -= engine.GetModule<TimeModule>().GetDeltatime().count() * 1e-3;
```
MuskettaMan, line 80: what is this scaling? ms to s?

MuskettaMan: because if GetDeltaTime returns a std::chrono time point, you could also use a std::chrono::duration_cast\<std::chrono::seconds\>()

shaders/particles/simulate.comp
```diff
@@ -52,7 +52,7 @@ void main()
         ParticleInstance instance;
         instance.position = particle.position;
         instance.materialIndex = particle.materialIndex;
-        instance.size = particle.size;
+        instance.size = particle.size.xy;
```
MuskettaMan, line 55: why change them to vec3's if you only use the vec2 part?

MuskettaMan (commented):
Some questions, but looks good overall, still need to do testing

engine/particles/private/particle_module.cpp
lyvanettinger: for some reason it's not guaranteed. I'm guessing newly created entities have their world matrix added after the particle module's tick since when I try to grab the world matrix component directly it just crashes

engine/particles/private/particle_module.cpp
lyvanettinger: yeah it's ms to s. I'll try casting

shaders/particles/simulate.comp
lyvanettinger: there's no implementation for it yet but I plan to add size scaling. just like the rotation scaling, they can shrink overtime for example. The third float is planned to be used as a 'velocity' for the scaling.

Not sure why I put this in this PR though xd I forgot about it halfway I think. Implementing functionality for this is added to a task in codecks though

PR #103, opened 2024-12-10
Features/ssao

dumitrurares14:
### Description

Added screen space ambient occlusion. We have a new graphics pipeline called ssao_pipeline that generates an AO texture that is sampled during the lighting pipeline to add on top of the final result.

Notes:

- There were some performance issues with this feature when you get too close to geometry, I tried to tackle it as best as possible without too much visual impact when you get too close to objects. Keep in mind that the effect might look different in those scenarios and is completely intended to avoid too much performance cost.
- SSAO is far from reality so don't expect to have realistic results. It looks nice, not realistic.
- For the moment this is a bit noisy; On top of the texture and other effects, it is not that noticeable and we can fix it by blurring it with the Gausian blur pipeline later. Don't complain about the noise, it's ok for the moment and some people even suggested it might even be nice for the artistic style..

BEFORE:
image

AFTER:
image

### Issues

https://bubonic-brotherhood.codecks.io/decks/8-graphics/card/1ak-ssao-user-story 
https://bubonic-brotherhood.codecks.io/card/1f8-ssao-shader 
https://bubonic-brotherhood.codecks.io/card/1f7-ssao-pipeline 

### Test criteria

- [ ] See if the output is something that we expect.
- [ ] Let me know if performance is impacted on different PCs. For me this works flawlessly
- [ ] Someone please test this on the steam deck.


engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
...
+            randomFloats(generator) * 2.0 - 1.0,
+            0.0f,
+            0.0f); // padding
+        ssaoNoise.push_back(noise);
+    }
+
+    auto resources { _context->Resources() };
+    auto cmdBuffer = SingleTimeCommands(_context->VulkanContext());
+
+    bblog::warn("{} {} {}", ssaoKernel[0].x, ssaoKernel[0].y, ssaoKernel[0].z);
```
mmzala, line 148: Delete unneeded log

engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
...
+    vk::PhysicalDeviceProperties properties {};
+    _context->VulkanContext()->PhysicalDevice().getProperties(&properties);
+
+    CreateBuffers();
+    CreateDescriptorSetLayouts();
+    CreateDescriptorSets();
+    CreatePipeline();
+}
+
+void SSAOPipeline::RecordCommands(vk::CommandBuffer commandBuffer, uint32_t currentFrame, MAYBE_UNUSED const RenderSceneDescription& scene) // NOLINT(*-include-cleaner)
```
mmzala, line 37: `NOLINT(*-include-cleaner)` comment can be deleted

engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
...
+        .SetData(std::move(byteData))
+        .SetFlags(vk::ImageUsageFlagBits::eSampled)
+        .SetFormat(vk::Format::eR32G32B32A32Sfloat);
+
+    noiseImage.isHDR = true;
+
+    _ssaoNoise = _context->Resources()->ImageResourceManager().Create(noiseImage);
+    _pushConstants.ssaoNoiseIndex = _ssaoNoise.Index();
+
+    // sizeof(glm::vec4);
```
mmzala, line 188: Delete unneeded comment

engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+        // scale magic to be more cluster near the actual fragment.
+        float scale = (float)i / 64.0;
+        scale = glm::mix(0.1f, 1.0f, scale * scale);
+        sample *= scale;
+        ssaoKernel.push_back(sample);
+    }
+
+    std::vector<glm::vec4> ssaoNoise;
+    for (unsigned int i = 0; i < 16; i++)
```
mmzala, line 135: Use `uin32_t`

engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
...
+        .SetDepthAttachmentFormat(vk::Format::eUndefined)
+        .BuildPipeline(_pipeline, _pipelineLayout);
+}
+void SSAOPipeline::CreateBuffers()
+{
+    // c++ side first
+    std::uniform_real_distribution<float> randomFloats(0.0, 1.0); // random floats between [0.0, 1.0]
+    std::default_random_engine generator;
+    std::vector<glm::vec4> ssaoKernel;
+    for (unsigned int i = 0; i < 64; ++i)
```
mmzala, line 117: Use `uin32_t`

engine/renderer/public/pipelines/ssao_pipeline.hpp
```diff
@@ -0,0 +1,66 @@
...
+    vk::DescriptorSetLayout _descriptorSetLayout;
+    vk::DescriptorSet _descriptorSet;
+
+    std::shared_ptr<GraphicsContext> _context;
+    const GBuffers& _gBuffers;
+    const ResourceHandle<GPUImage> _ssaoTarget;
+
+    vk::PipelineLayout _pipelineLayout;
+    vk::Pipeline _pipeline;
+    const CameraResource& _camera;
```
mmzala, line 63: Don't store the camera in the pipeline, you can get the descriptor set during the command recording, by getting the main camera through the GPUScene

engine/renderer/private/renderer.cpp
```diff
@@ -114,13 +117,23 @@ Renderer::Renderer(ApplicationModule& application, Viewport& viewport, const std
         .SetDebugLabelColor(glm::vec3 { 0.0f, 1.0f, 1.0f })
         .AddOutput(_gBuffers->Shadow(), FrameGraphResourceType::eAttachment);
 
+    FrameGraphNodeCreation ssaoPass { *_ssaoPipeline };
+    ssaoPass.SetName("SSAO pass")
+        .SetDebugLabelColor(glm::vec3(0.87f))
+        .AddInput(_gBuffers->Attachments()[0], FrameGraphResourceType::eTexture)
+        .AddInput(_gBuffers->Attachments()[1], FrameGraphResourceType::eTexture)
+        .AddInput(_gBuffers->Attachments()[2], FrameGraphResourceType::eTexture)
+        .AddInput(_gBuffers->Attachments()[3], FrameGraphResourceType::eTexture)
```
mmzala, line 126: The pass only uses position and normal buffers, so it should be specified as so here as well, otherwise the frame graph will think it has to put a memory barrier here for the other attachments here as well

engine/renderer/public/renderer.hpp
```diff
@@ -48,6 +49,7 @@ class Renderer
     std::shared_ptr<GraphicsContext> GetContext() const { return _context; }
     DebugPipeline& GetDebugPipeline() const { return *_debugPipeline; }
     BloomSettings& GetBloomSettings() { return *_bloomSettings; }
+    SSAOPipeline& GetSSAOPipeline() const { return *_ssaoPipeline; }
```
mmzala, line 52: SSAO Settings should be separate from the pass itself. There should be no need to get the pass outside of the renderer. Should be similar to something `BloomSettings`. But you can let this be for now, as we need to rework the renderer settings anyway.

engine/source/editor.cpp
```diff
@@ -29,6 +29,8 @@
 #include <fstream>
 #include <imgui/misc/cpp/imgui_stdlib.h>
 
+#include <pipelines/ssao_pipeline.hpp>
```
mmzala, line 32: Should use `""`, instead of `<>`, as the latter is meant for system includes

game/engine_api.wren
mmzala, line 1: Should not be included in the PR

shaders/ssao.frag
```diff
@@ -0,0 +1,92 @@
...
+    Camera camera;
+};
+
+layout (set = 1, binding = 0) buffer SampleKernel { vec4 samples[]; } uSampleKernel;
+
+layout (location = 0) in vec2 texCoords;
+layout (location = 0) out vec4 outColor;
+
+
+const int kernelSize = 64;
```
mmzala, line 32: Shouldn't this be a parameter that you pass to the shader?

engine/renderer/public/pipelines/lighting_pipeline.hpp
```diff
@@ -43,4 +44,5 @@ class LightingPipeline final : public FrameGraphRenderPass
     vk::Pipeline _pipeline;
 
     const BloomSettings& _bloomSettings;
+    const ResourceHandle<GPUImage> _ssaoTarget;
```
mmzala, line 47: No need to store this here, as it is only used to get the index to the push constants

engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    commandBuffer.bindPipeline(vk::PipelineBindPoint::eGraphics, _pipeline);
+
+    commandBuffer.pushConstants<PushConstants>(_pipelineLayout, vk::ShaderStageFlagBits::eFragment, 0, _pushConstants);
+
+    commandBuffer.bindDescriptorSets(vk::PipelineBindPoint::eGraphics, _pipelineLayout, 0, { _context->BindlessSet() }, {});
+    commandBuffer.bindDescriptorSets(vk::PipelineBindPoint::eGraphics, _pipelineLayout, 1, { _descriptorSet }, {});
+    commandBuffer.bindDescriptorSets(vk::PipelineBindPoint::eGraphics, _pipelineLayout, 2, _camera.DescriptorSet(currentFrame), {});
+
+    // To add the descirpot sets for the kernel and noise buffers
```
mmzala, line 69: Delete unneeded comment

engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "gpu_scene.hpp"
+#include "graphics_context.hpp"
+#include "graphics_resources.hpp"
+#include "pipeline_builder.hpp"
+#include "resource_management/image_resource_manager.hpp"
+#include "shaders/shader_loader.hpp"
+#include "vulkan_context.hpp"
+
+#include <random>
+#include <resource_management/buffer_resource_manager.hpp>
```
mmzala, line 15: Use `""` for our headers

engine/renderer/private/pipelines/ssao_pipeline.cpp
```diff
@@ -0,0 +1,233 @@
+﻿#include "pipelines/ssao_pipeline.hpp"
+
+#include "../vulkan_helper.hpp"
```
mmzala, line 3: Why go a directory back? You can just include it without doing so

mmzala (changes requested):
Good work, I tested and it looks nice. I left some comments to clean up the code

engine/renderer/private/pipelines/ssao_pipeline.cpp
dumitrurares14: My bad, clion auto-complete probably incldued it like this

engine/renderer/private/pipelines/ssao_pipeline.cpp
dumitrurares14: Done 💪

shaders/ssao.frag
dumitrurares14: I chose to left it hard-coded 64, since the kernel is generated and cached on the CPU side with that size anyway and I don't want to regenerate it later.

engine/renderer/public/pipelines/lighting_pipeline.hpp
dumitrurares14: Removed 🫡

engine/renderer/private/renderer.cpp
mmzala: Make sure to remove targets 0 and 2 as they are not used here

engine/renderer/public/pipelines/ssao_pipeline.hpp
mmzala: Make sure to do this as well

game/engine_api.wren
mmzala, line 1: Remove this file from the PR

shaders/ssao.frag
mmzala: Alright, a nitpick but can you fix the formatting, it bothers me that the variable is directly above the main function XD

mmzala (commented):
There are still a couple of things that need to be changed before merging, since some of them we don't want to have on main

mmzala (approved):
Nice work!

PR #104, opened 2024-12-10
Wren Hot Reloading

santiberna:
### Description

This pull request add the Inspector Module, the functionality to reset the scripting module and reload the main script and a button to do this on the main toolbar.

The inspector module separates most of the imgui functionality into its own module.
Wren VM hot reloading is relatively fast.

### Issues

[Codecks Issue](https://bubonic-brotherhood.codecks.io/card/1c3-wren-game-main-entry-function)

### Test criteria

You can test the Hot Reloading by playing with the initial starting position of the player in wren. It should also be tested if ImGui initialization and shutdown is handled correctly in the new module


engine/inspector/private/inspector_module.cpp
```diff
@@ -0,0 +1,89 @@
...
+    if (auto* ptr = engine.GetModuleSafe<RendererModule>())
+    {
+        ptr->GetRenderer()->FlushCommands();
+    }
+    _editor.reset();
+}
+
+void InspectorModule::Tick(Engine& engine)
+{
+    _imguiBackend->NewFrame();
```
luuk234, line 54: i would move all the other editor logic to here as well probably

luuk234 (approved):

engine/inspector/private/inspector_module.cpp
santiberna: I believe I got all of the logic, what am I missing something?

engine/include/old_engine.hpp
```diff
@@ -27,13 +26,9 @@ class OldEngine : public ModuleInterface
     // std::unique_ptr<ThreadPool> _threadPool;
     // std::unique_ptr<AssetManager> _AssetManager;
 
-    std::unique_ptr<Editor> _editor;
     ECSModule* _ecs;
     glm::ivec2 _lastMousePos {};
     std::chrono::time_point<std::chrono::high_resolution_clock> _lastFrameTime;
-    PerformanceTracker _performanceTracker;
-
-    std::unique_ptr<PhysicsModule> _physicsModule;
```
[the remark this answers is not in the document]
santiberna: will check

engine/inspector/private/inspector_module.cpp
luuk234: well the editor draw is basically what draws all our imgui windows right, bit of an odd distinction to put the imgui titlebar in here but the rest of the editor logic in said function

santiberna: Oh yeah you are right... I just thought moving all the editor code outside of that class could take some time and make the inspector module. I will add a TODO here

luuk234: alright good enough, i also mentioned for the future since this is out of scope for this pr

PR #107, opened 2024-12-11
Feature/fxaa

dumitrurares14:
### Description

Added FXAA pipeline which takes as input the tone mapping results. The UI will now receive the FXAA target as input to chain things together nicely.

Implementation based on this:
https://blog.simonrodriguez.fr/articles/2016/07/implementing_fxaa.html#fnref3 

Notes:
- Performance should not be an issue with this feature. Nsight measures a frame of around 0.17-0.20 ms for this pipeline and this is with pretty "High" settings 
image

- The effect is not that modifiable, settings that are exposed are offering pretty minimal changes, so keep in mind that if we want better results than what the current default settings are, it means we need a different AA method.

Here are some comparison shots:

Before:
image

After:
image

### Issues

https://bubonic-brotherhood.codecks.io/decks/8-graphics/card/193-anti-aliasing-user-story
https://bubonic-brotherhood.codecks.io/decks/8-graphics/card/1f9-fxaa-pipeline
https://bubonic-brotherhood.codecks.io/decks/8-graphics/card/1fa-fxaa-shader

### Test criteria

- [x] Let me know what do you think about the results
- [ ] I would love some feedback on the pipeline implementation. I am using only bindless and I am not sure if this is good or not.


engine/inspector/private/editor.cpp
```diff
@@ -16,15 +16,20 @@
 #include "log.hpp"
 #include "menus/performance_tracker.hpp"
 #include "pipelines/ssao_pipeline.hpp"
+#include "model_loader.hpp"
+#include "physics_module.hpp"
```
LeoHellmig, line 20: Why are we adding the physics_module in this fxaa pr

engine/renderer/public/pipelines/fxaa_pipeline.hpp
```diff
@@ -0,0 +1,55 @@
...
+    ~FXAAPipeline() final;
+
+    void RecordCommands(vk::CommandBuffer commandBuffer, uint32_t currentFrame, const RenderSceneDescription& scene) final;
+
+    float& GetEdgeTreshholdMin() { return _pushConstants.edgeThresholdMin; }
+    float& GetEdgeTreshholdMax() { return _pushConstants.edgeThresholdMax; }
+    float& GetSubPixelQuality() { return _pushConstants.subPixelQuality; }
+    int32_t& GetIterations() { return _pushConstants.iterations; }
+    bool& GetEnableFXAA() { return _pushConstants.enableFXAA; }
+
```
LeoHellmig, line 28: Don't write getter functions like this, either make the variables public or add the classes that need access friend classes

engine/renderer/public/pipelines/fxaa_pipeline.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "gbuffers.hpp"
+
+#include <cstdint>
+#include <memory>
+
+class BloomSettings;
+class GraphicsContext;
+class CameraResource;
+struct GPUImage;
+struct RenderSceneDescription;
```
LeoHellmig, line 13: Forward declarations for BloomSettings and CameraResource are not being used

engine/renderer/public/pipelines/fxaa_pipeline.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+class FXAAPipeline final : public FrameGraphRenderPass
+{
+public:
+    FXAAPipeline(const std::shared_ptr<GraphicsContext>& context, const GBuffers& gBuffers, const ResourceHandle<GPUImage>& fxaaTarget, const ResourceHandle<GPUImage>& sourceTarget);
+    ~FXAAPipeline() final;
+
+    void RecordCommands(vk::CommandBuffer commandBuffer, uint32_t currentFrame, const RenderSceneDescription& scene) final;
```
LeoHellmig, line 21: nitpick The "final" keyword here is not needed, do specify the functions as override

engine/renderer/public/pipelines/fxaa_pipeline.hpp
```diff
@@ -0,0 +1,55 @@
+﻿#pragma once
+
+#include "frame_graph.hpp"
+#include "gbuffers.hpp"
+
```
LeoHellmig, line 5: #include "gbuffers.hpp" is not needed here

engine/renderer/private/pipelines/fxaa_pipeline.cpp
```diff
@@ -0,0 +1,96 @@
+﻿#include "pipelines/fxaa_pipeline.hpp"
+
+#include "graphics_context.hpp"
+#include "graphics_resources.hpp"
+#include "pipeline_builder.hpp"
+#include "resource_management/image_resource_manager.hpp"
+#include "shaders/shader_loader.hpp"
+#include "vulkan_context.hpp"
+
+#include <random>
```
LeoHellmig, line 10: Why are you including <random>

engine/renderer/private/pipelines/fxaa_pipeline.cpp
```diff
@@ -0,0 +1,96 @@
...
+FXAAPipeline::FXAAPipeline(const std::shared_ptr<GraphicsContext>& context, const GBuffers& gBuffers, const ResourceHandle<GPUImage>& fxaaTarget, const ResourceHandle<GPUImage>& sourceTarget)
+    : _pushConstants()
+    , _context(context)
+    , _gBuffers(gBuffers)
+    , _fxaaTarget(fxaaTarget)
+    , _source(sourceTarget)
+{
+
+    vk::PhysicalDeviceProperties properties {};
+    _context->VulkanContext()->PhysicalDevice().getProperties(&properties);
```
LeoHellmig, line 22: Are these properties ever used?

LeoHellmig (changes requested):
The result looks very good, I can't really say anything about the pipeline implementation except some nitpicks here and there which I would like to see fixed

engine/renderer/public/pipelines/fxaa_pipeline.hpp
dumitrurares14: Fixed

engine/renderer/public/pipelines/fxaa_pipeline.hpp
dumitrurares14: I dont know about this one. Nobody said anything when I did the same thing in the SSAO pr. If its a big problem i can change it. The rest of the things are pushed now 🫡

LeoHellmig: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-get 
Its fine, but you should read this sometime

LeoHellmig (approved):

MuskettaMan (approved):
Nice work! just a few minor nitpicks

PR #108, opened 2024-12-14
Tracy usage

LeoHellmig:
### Description

Using tracy in more parts of the engine.
Now that we are no longer using MinGW anymore, Tracy profiler actually works again. I have added some zones in our codebase which covers most of our cpu time roughly. 

This is useful to us as a team, because tracy allows us to find time is spent and memory is allocated and deallocated. 

image
Example trace where you can see different zones.

### Issues

None

### Test criteria

None
But do use tracy to profile your code and the application in general regularly


MuskettaMan (approved):
Looks good! Only question i have, cant we integrate profiling in the module interface abstraction? When all modules are ticked we can add profiling for them from there. Meaning that all modules we add automatically get registered in tracy. What do you think?

LeoHellmig:
> Looks good! Only question i have, cant we integrate profiling in the module interface abstraction? When all modules are ticked we can add profiling for them from there. Meaning that all modules we add automatically get registered in tracy. What do you think?

This sounds like a great Idea, I'll implement this by getting the name of the typeI I believe clang has support for that

PR #109, opened 2024-12-14
Fix/nan normal

MuskettaMan:
### Description

Fixed several bugs in the renderer:
- Nan issues in the normal buffer
  - Fixed by how we process tangents and test for incorrect ones.
- Shadows looking pixelated, and ssao looking bad
  - Added more precision in the position buffer  
- Bloom bleeding through on the edges of the screen
  - Added a border color to the sampler to prevent this
- Glitching mesh
  - Was due to a very small scaling value, fixed this in the model directly

### Issues

https://bubonic-brotherhood.codecks.io/card/1fy-meshes-in-the-dark-fantasy-asset-have-nan-normals

### Test criteria

- [ ] Look at the bugs and make sure they dont occur anymore


engine/renderer/private/gpu_scene.cpp
```diff
@@ -115,6 +119,11 @@ void GPUScene::UpdateObjectInstancesData(uint32_t frameIndex)
         instances[count].materialIndex = mesh->material.Index();
         instances[count].boundingRadius = mesh->boundingRadius;
 
+        if (count == 8038)
```
VengarvdWal: If we were to load a model before this one, this would no longer be correct, right?

MuskettaMan: yes correct, this was just some a test condition for my breakpoint that i forgot to remove

mmzala (approved):
LGTM

PR #112, opened 2024-12-16
Feature/compute reflect

MuskettaMan:
### Description

Implemented a PipelineBuilder class that supports compute pipelines. This involved creating a inheritance hierarchy for pipeline builder that has an implementation for graphics and compute. Since they are similar but still effectively different.

There was also a separate issue that popped up where the creation of DSLs is difficult to manage when two interface blocks are in glsl that are the same DSL but have to be named differently. For now this is handled by using a underscore as an delimiter. Later this ideally should be fixed by generating DSLs during compile time.

There are also some changes regarding the reflect.py script. This isn't used yet, it was just me messing around with it.

I also updated the cathedral glb file to fix another scaling issue that glitched out some things.

### Issues

https://bubonic-brotherhood.codecks.io/card/1gz-compute-pipeline-builder

### Test criteria

- [x] Make sure the scene still compiles as before and logs no warnings or errors


lyvanettinger (approved):
Great work! I have no comments on the code itself and have tested the branch and it still works all good on my end

PR #114, opened 2024-12-17
Audio system and 3D sounds/events

LeoHellmig:
### Description

This pull request adds the following features to the engine:
- ``AudioListenerComponent`` A component tag which specifies what entity is currently listening to sound. Its 3D attributes like position and velocity are sent to FMOD for spatialization.
- ``AudioEmitterComponent`` A component which holds IDs to currently playing sounds or events. In the ``AudioSystem`` any 3D sounds or events will be updated to the position of the emitter that holds a reference to them. 
- ``AudioSystem`` 
     - The audio system manages the audio components in the ECS. Telling the audio module where our current listener is. 
     - Removing any sounds or events if they have stopped playing from emitters. 
     - Updating the FMOD 3D attributes for 3D sounds and events. 
     - Velocity for 3D attributes can be get from the physics module
     - Audio module inspection 
            - Showing amount of currently loaded sounds and banks.
            - Showing amount of currently playing sounds and events.
            - Spectrum analyser for the mastergroup output.
 - Upgraded handle system, where types make it clear what a function returns or expects
 - Basic debug lines for sounds and events
 - Examples of audio usage in the engine
 - Some new transform helpers

https://github.com/user-attachments/assets/1ac89b86-8e57-4bc2-a61c-4e3433872d6c

### Issues

https://bubonic-brotherhood.codecks.io/card/1ev-emitter-functionality 
https://bubonic-brotherhood.codecks.io/card/1ec-audio-system

### Test criteria

- [ ] Try building the project and see if the sound works as expected


engine/audio/private/audio_system.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+AudioSystem::AudioSystem(ECSModule& ecs, AudioModule& audioModule)
+    : _ecs(ecs)
+    , _audioModule(audioModule)
+{
+}
+void AudioSystem::Update(ECSModule& ecs, MAYBE_UNUSED float dt)
+{
+    ZoneScoped;
+    const auto& listenerView = ecs.GetRegistry().view<AudioListenerComponent>();
```
santiberna, line 25: Having an Audio Listener component might be a little useless since only one entity can really have this component. Either way, it works for tagging the player entity

engine/audio/private/audio_system.cpp
```diff
@@ -0,0 +1,159 @@
...
+        const glm::vec3 position = TransformHelpers::GetWorldPosition(ecs.GetRegistry(), entity);
+        glm::vec3 velocity = glm::vec3(0.f, 0.f, 0.f);
+
+        const glm::quat rotation = TransformHelpers::GetWorldRotation(ecs.GetRegistry(), entity);
+        const glm::vec3 forward = glm::normalize(rotation * glm::vec3(0.f, -1.f, 0.f));
+        const glm::vec3 up = glm::normalize(rotation * glm::vec3(0.f, 0.f, 1.f));
+
+        if (RigidbodyComponent* rigidBody = ecs.GetRegistry().try_get<RigidbodyComponent>(entity))
+        {
+            velocity = ToGLMVec3(_audioModule._physics->bodyInterface->GetLinearVelocity(rigidBody->bodyID));
```
santiberna, line 81: Oh ok I get why we need the physics module here

engine/audio/private/audio_system.cpp
```diff
@@ -0,0 +1,159 @@
...
+        }
+        // Update 3D position of sounds
+        for (auto soundInstance : emitter._soundIds)
```
santiberna, line 84: Why are we handling both events and sounds at the same time? Shouldn't this logic only work with events that then propagate to the sounds?

engine/source/engine.cpp
```diff
@@ -33,6 +35,9 @@
 
 #include <time_module.hpp>
 
+SoundInfo si;
```
santiberna, line 38: So how are sounds tracked by the engine? It might be more intuitive to pass the sound infos to the audio module so they are loaded, and the user gets a SoundHandle returned. This also avoids someone loading the sound and then changing the path or any properties after loading.

The sound handle is then used for calling sounds and updating 3D values and such

santiberna (changes requested):
Wow this is a lot of work for getting the 3D sounds working. I think it will add a lot in terms of immersion for our game!

I did highlight some doubts and issues I see in your code. My biggest problem is mainly understanding what are the differences between events and sounds, because it seems like we are using both but I feel like we should only use events for gameplay logic. More comments below

santiberna (commented):
Just removing the requested changes since I will be away

engine/audio/private/audio_module.cpp
```diff
@@ -48,81 +62,105 @@ void AudioModule::Tick(MAYBE_UNUSED Engine& engine)
...
     FMOD_CHECKRESULT(FMOD_System_CreateSound(_coreSystem, soundInfo.path.data(), mode, nullptr, &sound));
 
     _sounds[hash] = sound;
+    _soundInfos[hash] = &soundInfo;
 }
-void AudioModule::PlaySFX(SoundInfo& soundInfo)
+SoundInfo& AudioModule::GetSFX(const std::string_view path)
+{
+    const SoundInfoID hash = std::hash<std::string_view> {}(path);
+    const auto it = std::ranges::find_if(_soundInfos, [&](const std::unordered_map<SoundInfoID, SoundInfo*>::value_type& vt)
```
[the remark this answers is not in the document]
LeoHellmig: This is indeed a lot simpler, thank you

engine/audio/private/audio_system.cpp
LeoHellmig: In the case of fmod, sounds and events are 2 different audio types.
It doesn't have anything to do with an event system.
Sounds are simple samples loaded from .mp3 .ogg .wav files.
While FMOD events are their own more powerful audio type including many variables and possible special objects.

engine/audio/private/audio_system.cpp
LeoHellmig: It is indeed just a tag for the audio system to know what entity is listening, it could possibly later be expanded

engine/source/engine.cpp
LeoHellmig: This is a better idea yea, I have implemented it thank you  👍

engine/audio/private/audio_common.cpp
```diff
@@ -0,0 +1 @@
+#include "audio_common.hpp"
```
santiberna, line 1: Why does this file exist?

engine/audio/private/audio_system.cpp
santiberna: Ok I got it!

santiberna (approved):
Good job simplifying the system with returning handles and using unordered maps of strings instead of ids!

PR #116, opened 2024-12-17
Feature/steam public build

mmzala:
### Description

- Adds scripts to upload the distribution builds to the public depot on Steam
- Renamed the development depot upload scripts to match their configuration

The public configuration can cannot be set active on Steam automatically, unlike the development configuration, it needs to be set active manually on the Steamworks website. But it can be sent to the Steam servers, that's what this PR does.

### Issues

[Codecks card](https://bubonic-brotherhood.codecks.io/milestones/run/44/card/1g3-steam-build-depot-for-public-releases).

### Test criteria

Uploading build:

1. Run the `package_and_upload_steam_public.bat`
2. Fill in the Steam login info
3. The build should have uploaded to Steam successfully
4. Go to the Steamworks website
5. Go to the newly uploaded build and set it active for the `default` branch
6. Open Steam and run the game on the default branch


santiberna (commented):
Make sure that all of those ids are safe to place publicly on GitHub. If not, maybe we should move these scripts to Perforce since we are going to start working with it for assets as well and we can maybe store more sensitive info there.

Didn't test, its probably better someone there does it so you can help in person in case something breaks.

MuskettaMan (approved):

PR #117, opened 2024-12-18
Features/mesh shapes

dumitrurares14:
### Description

Added support for complex hierarchies with jolt. 
Added mesh and convex hull colliders for entities with static mesh component
Added function to load only collision data as gltf using either mesh or convex hull shapes

image
image

### Issues

https://bubonic-brotherhood.codecks.io/card/1g1-mesh-shapes-for-colliders
https://bubonic-brotherhood.codecks.io/decks/7-engine/card/1g5-sync-physics-and-rendering-with-hierarchy

### Test criteria

- [ ] Play around in the editor to see how things work
- [ ] There should still be some minor issues that can be fix with early returns in certain scenarios.
- [ ] There is this big issue where dynamic bodies that have non uniform scaling will reset the mesh scale to 1.0 during simulation because jolt world transform does not hold scale. If you have any ideas how to fix this please help.

### DIFF
--- a/engine/ecs/public/components/transform_component.hpp
+++ b/engine/ecs/public/components/transform_component.hpp
@@ -20,7 +20,10 @@ private:
     friend class TransformHelpers;
     friend class Editor;
 };
-
+struct ToBeUpdated
+{
+    int padding {};
+};
 namespace EnttEditor
 {
 template <>
--- a/engine/physics/private/systems/physics_system.cpp
+++ b/engine/physics/private/systems/physics_system.cpp
@@ -14,6 +14,9 @@
 #include "renderer_module.hpp"
 #include "resource_management/mesh_resource_manager.hpp"
 
+#include <model_loader.hpp>
+#include <renderer.hpp>
+
 PhysicsSystem::PhysicsSystem(Engine& engine, ECSModule& ecs, PhysicsModule& physicsModule)
     : engine(engine)
     , _ecs(ecs)
@@ -24,7 +27,7 @@ PhysicsSystem::PhysicsSystem(Engine& engine, ECSModule& ecs, PhysicsModule& phys
 void PhysicsSystem::InitializePhysicsColliders()
 {
     const auto view = _ecs.GetRegistry().view<StaticMeshComponent, TransformComponent>();
- 
+
     for (const auto entity : view)
     {
         StaticMeshComponent& meshComponent = view.get<StaticMeshComponent>(entity);
@@ -65,6 +68,43 @@ void PhysicsSystem::InitializePhysicsColliders()
         _ecs.GetRegistry().emplace_or_replace<UpdateMeshAndPhysics>(entity);
     }
 }
+
+void PhysicsSystem::CreateMeshCollision(const std::string& path)
+{
+    const CPUModel models = engine.GetModule<RendererModule>().GetRenderer().get()->GetModelLoader().ExtractModelFromGltfFile(path);
+
+    for (auto mesh : models.meshes)
+    {
+        JPH::VertexList vertices;
+        JPH::IndexedTriangleList triangles;
+
+        // set verticies
+        for (auto vertex : mesh.vertices)
+        {
+            vertices.push_back(JPH::Float3(vertex.position.x, vertex.position.y, vertex.position.z));
+        }
+
+        // set trinagles
+        for (size_t i = 0; i + 2 < mesh.indices.size(); i += 3)
+        {
+            JPH::IndexedTriangle tri;
+            tri.mIdx[0] = mesh.indices[i + 0];
+            tri.mIdx[1] = mesh.indices[i + 1];
+            tri.mIdx[2] = mesh.indices[i + 2];
+
+            triangles.push_back(tri);
+        }
+
+        RigidbodyComponent rb(*_physicsModule.bodyInterface, entt::null, glm::vec3(0.0f, 0.0f, 0.0f), vertices, triangles);
+
+        entt::entity entity = _ecs.GetRegistry().create();
+        NameComponent node;
+        node.name = "Mesh collider Entity";
+        _ecs.GetRegistry().emplace<NameComponent>(entity, node);
+        _ecs.GetRegistry().emplace<RigidbodyComponent>(entity, rb);
+    }
+}
+
 void PhysicsSystem::CleanUp()
 {
     const auto toDestroy = _ecs.GetRegistry().view<DeleteTag, RigidbodyComponent>();
@@ -178,6 +218,11 @@ void PhysicsSystem::Inspect()
         node.name = "Plane Entity";
         _ecs.GetRegistry().emplace<RigidbodyComponent>(entity, newRigidBody);
         _ecs.GetRegistry().emplace<NameComponent>(entity, node);
+    }
+
+    if (ImGui::Button("Create mesh collider"))
+    {
+        CreateMeshCollision("assets/models/collision_test.glb");
     }
 
     if (ImGui::Button("Clear Physics Entities"))
--- a/engine/physics/private/systems/physics_system.cpp
+++ b/engine/physics/private/systems/physics_system.cpp
@@ -73,7 +73,7 @@ void PhysicsSystem::InitializePhysicsColliders()
 }
 entt::entity PhysicsSystem::LoadNodeRecursive(const CPUModel& models, ECSModule& ecs,
     uint32_t currentNodeIndex,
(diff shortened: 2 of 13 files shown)


engine/physics/private/systems/physics_system.cpp
```diff
@@ -89,42 +222,49 @@ void PhysicsSystem::Update(MAYBE_UNUSED ECSModule& ecs, MAYBE_UNUSED float delta
...
+        RelationshipComponent& relationship = _ecs.GetRegistry().get<RelationshipComponent>(entity);
+        if (relationship.parent != entt::null)
+            RelationshipHelpers::DetachChild(_ecs.GetRegistry(), relationship.parent, entity);
+
+        // Crazy jolt stuff that I dont like but it works to set the proper scale
+        JPH::BodyLockWrite lock(_physicsModule.physicsSystem->GetBodyLockInterface(), rb.bodyID);
+        if (lock.Succeeded())
+        {
+            JPH::Body& body = lock.GetBody();
+            const JPH::ScaledShape* scaled_shape = static_cast<const JPH::ScaledShape*>(body.GetShape());
```
santiberna, line 255: Probably use dynamic cast since I am not sure how error prone this is. Are you sure all of the shapes we create derive from ScaledShape?

engine/physics/private/systems/physics_system.cpp
```diff
@@ -195,16 +345,13 @@ void PhysicsSystem::Inspect()
...
-    if (_ecs.GetRegistry().valid(entity))
-        _ecs.GetRegistry().emplace_or_replace<UpdateMeshAndPhysics>(entity);
 
     ImGui::PushID(&rb.bodyID);
     JPH::Vec3 position = _physicsModule.bodyInterface->GetPosition(rb.bodyID);
     float pos[3] = { position.GetX(), position.GetY(), position.GetZ() };
     if (ImGui::DragFloat3("Position", pos, 0.1f))
     {
-        _physicsModule.bodyInterface->SetPosition(rb.bodyID, JPH::Vec3(pos[0], pos[1], pos[2]), JPH::EActivation::Activate);
+        _physicsModule.physicsSystem->GetBodyInterfaceNoLock().SetPosition(rb.bodyID, JPH::Vec3(pos[0], pos[1], pos[2]), JPH::EActivation::Activate);
```
santiberna, line 354: Why not use the locked interface?

engine/physics/public/components/rigidbody_component.hpp
```diff
@@ -50,12 +53,77 @@ struct RigidbodyComponent
...
+        // lets save thes shape reference
+        shape = bodyCreationSettings.GetShape();
         bodyID = bodyInterface.CreateAndAddBody(bodyCreationSettings, JPH::EActivation::Activate);
 
         // set the owner entity so we can querry it later from physics ohbejcts if needed
         bodyInterface.SetUserData(bodyID, static_cast<uintptr_t>(ownerEntity));
     }
 
+    // for mesh collisions
+    RigidbodyComponent(JPH::BodyInterface& bodyInterface, entt::entity ownerEntity, glm::vec3 position, JPH::VertexList& vertices, JPH::IndexedTriangleList& triangles)
```
santiberna, line 65: All of this shape configuration and creation should be done externally and then passed to the RigidBody component. I don't think components should have so much complex logic for initialization and you should probably move it to the Physics Module or System

engine/source/engine.cpp
```diff
@@ -49,14 +49,15 @@ ModuleTickOrder OldEngine::Init(Engine& engine)
...
         //"assets/models/Adventure.glb",
         //"assets/models/DamagedHelmet.glb",
         //"assets/models/CathedralGLB_GLTF.glb",
-        // "assets/models/Terrain/scene.gltf",
-        "assets/models/ABeautifulGame/ABeautifulGame.gltf",
-        //"assets/models/MetalRoughSpheres.glb"
+        //"assets/models/Terrain/scene.gltf",
+        //"assets/models/ABeautifulGame/ABeautifulGame.gltf",
+        //"assets/models/MetalRoughSpheres.glb",
+        "assets/models/monkey.gltf"
```
santiberna, line 60: Switch these back to default when you merge

engine/source/engine.cpp
```diff
@@ -68,8 +69,11 @@ ModuleTickOrder OldEngine::Init(Engine& engine)
 
     _ecs = &engine.GetModule<ECSModule>();
 
-    SceneLoading::LoadModelIntoECSAsHierarchy(*_ecs, *modelResourceManager.Access(models[0].second), models[0].first.hierarchy, models[0].first.animation);
-    SceneLoading::LoadModelIntoECSAsHierarchy(*_ecs, *modelResourceManager.Access(models[1].second), models[1].first.hierarchy, models[1].first.animation);
+    SceneLoading::LoadModelIntoECSAsHierarchy(*_ecs, *modelResourceManager.Access(models[0].second), models[0].first, models[0].first.hierarchy, models[0].first.animation);
+
+    // to add back before merge
```
santiberna, line 74: Dont forget to add back before merge

santiberna (changes requested):
Still a couple changes I need to see.,I also have some question so please answer below. Good work!

engine/physics/private/systems/physics_system.cpp
dumitrurares14: This was changed purely based on some documentation suggestions. If you use no lock the body wont lock to random position which can happen in some scenarios apparently or something like that. Jolt docs are very bad and I don't really have a better answer other than I think it's better like this. At the end of the day it doesn't really matter that much since it's used only in the editor in this scenario.

engine/ecs/public/components/transform_component.hpp
```diff
@@ -20,7 +20,10 @@ struct TransformComponent
     friend class TransformHelpers;
     friend class Editor;
 };
-
+struct ToBeUpdated
+{
+    int padding {};
```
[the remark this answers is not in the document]
dumitrurares14: This was removed

engine/physics/private/systems/physics_system.cpp
```diff
@@ -50,23 +107,75 @@ void PhysicsSystem::InitializePhysicsColliders()
...
+        rb = RigidbodyComponent(*_physicsModule.bodyInterface, entityToAttachTo, glm::vec3(0.0), vertices);
+    return rb;
+}
+void PhysicsSystem::CreateConvexHullCollision(const std::string& path)
+{
+    CPUModel models = _collisionLoader.get()->ExtractModelFromGltfFile(path);
+
+    LoadNodeRecursive(models, _ecs, models.hierarchy.root, models.hierarchy, entt::null, eCONVEXHULL);
+}
+void PhysicsSystem::CreateMeshCollision(const std::string& path)
```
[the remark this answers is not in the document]
dumitrurares14: Fixed 🫡Probably rushed it and I didn't noticed I made it like this

engine/physics/private/systems/physics_system.cpp
dumitrurares14: Using static cast is what they suggest in the documentation and their example. At a first glance it works with dynamic cast also but I have no idea if it can break anything later so I would rather keep it like they wanted. Check their example if you want:

https://github.com/jrouwe/JoltPhysics/blob/master/Samples/Tests/ScaledShapes/DynamicScaledShape.cpp

engine/physics/private/systems/physics_system.cpp
```diff
@@ -65,6 +70,142 @@ void PhysicsSystem::InitializePhysicsColliders()
         _ecs.GetRegistry().emplace_or_replace<UpdateMeshAndPhysics>(entity);
     }
 }
+entt::entity PhysicsSystem::LoadNodeRecursive(const CPUModel& models, ECSModule& ecs,
```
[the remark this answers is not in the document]
dumitrurares14: Fixed with an extra log as we discussed on discord

engine/physics/private/systems/physics_system.cpp
santiberna: Okay but... if dynamic_cast fails we get a crash. If static cast fails, we might have a bug that is almost impossible to detect

engine/physics/private/systems/physics_system.cpp
santiberna: Okay you can keep it, although the lock in this context refers to multithreading... weird

santiberna (approved):
Okay looks good!! We can merge now!

PR #118, opened 2024-12-18
Feature/audio wren bindings

LeoHellmig:
### Description

Wren bindings for audio
This pull request has wren bindings for basic audio loading and playing through scripts.
There is also a demo included, holding right trigger will play a shooting sound

### Issues

https://bubonic-brotherhood.codecks.io/card/1ek-wren-bindings-for-audio

### Test criteria

- [x] See if the project builds
- [x] Test if holding down the right trigger on a controller plays a shooting sound


engine/bindings/private/audio/audio_bindings.cpp
```diff
@@ -0,0 +1,88 @@
...
+    si.isLoop = isLoop;
+
+    self.LoadSFX(si);
+}
+
+SoundInstance PlaySFX(AudioModule& self, const std::string& path, const float volume)
+{
+    if (!self.isSFXLoaded(path))
+    {
+        bblog::error("Tried to play a sound that was not loaded: {0}", path);
```
santiberna, line 36: Shouldn't you return early in this function if the sound was not loaded? Wrenbindings support the use of optionals, which means for wren you can return a SoundInstance or None

santiberna (approved):
Tested and works great. Looks pretty clean! I really need to separate the other bindings into their own files because the way you did it made it very easy to read.

I am going to approve but there is a small comment you can consider in one of your bindings.

engine/bindings/private/audio/audio_bindings.cpp
LeoHellmig: Oh very interesting, I'll implement it. Thank you!

PR #119, opened 2024-12-19
Feature/multiple animations

MuskettaMan:
### Description

Adds functionality for improved loading of skeletal meshes, now works with more kinds of gltfs.
Adds a gun and arms models onto the camera, with animations.
Adds support for multiple animation clips.
Adds support for wren bindings into the animations system.
Adds a small demo 

### Issues

- https://bubonic-brotherhood.codecks.io/card/1ei-wren-skeleton-animation-bindings
- https://bubonic-brotherhood.codecks.io/card/1e4-interacting-with-skeletal-animation-system

### Test criteria

- [x] Make sure the gun animations work
- [x] Test the input controls, currently this is 'R' key for reload, and LMB for shooting
- [x] Review the interface that is exposed to wren, does it look usable or is missing something
- [x] Review how I expose my bindings in the binding module, perhaps this can be split up better, or use different technique for exposing them. Because right now Im doing it through funcExt, since its a wren component


santiberna (commented):
In general I think you did the wren bindings well, in terms of components it is a bit unavoidable to not use external functions because of the WrenComponent<T> template. If you feel like it, I would also like it if you split the math wren bindings to another file, to keep it tidy like Leo did with the audio module. 

Besides, I have some small notes in the PR, check below

dumitrurares14:
I see Santiago reviewed the wren stuff. As for the animation I couldn't find anything to pick on. The implementation and how they are laid out is elegant. Nice stuff 👍, I assume Santiago will approve after the separation of the math thing

engine/renderer/public/animation.hpp
```diff
@@ -41,24 +42,128 @@ struct AnimationSpline
...
+    {
+        activeAnimation = animationIndex;
+        animations[animationIndex].time = 0.0f;
+        animations[animationIndex].looping = looping;
+        animations[animationIndex].speed = speed;
+        animations[animationIndex].playbackOption = Animation::PlaybackOptions::ePlaying;
+    }
+    void Play(const std::string& name, float speed = 1.0f, bool looping = false)
+    {
+        auto it = std::find_if(animations.begin(), animations.end(), [&name](const auto& anim)
```
santiberna, line 97: Forgor to include algorithm

engine/renderer/public/animation.hpp
```diff
@@ -41,24 +42,128 @@ struct AnimationSpline
...
+};
+
+struct AnimationControlComponent
+{
+    std::vector<Animation> animations;
+    std::optional<uint32_t> activeAnimation { std::nullopt };
+
+    void PlayByIndex(uint32_t animationIndex, float speed = 1.0f, bool looping = false)
+    {
+        activeAnimation = animationIndex;
```
santiberna, line 89: Probably move all these function definitions to a cpp file

engine/renderer/public/components/animation_channel_component.hpp
```diff
@@ -5,11 +5,16 @@
...
     std::optional<AnimationSpline<Translation>> translation { std::nullopt };
     std::optional<AnimationSpline<Rotation>> rotation { std::nullopt };
     std::optional<AnimationSpline<Scale>> scaling { std::nullopt };
 };
+
+struct AnimationChannelComponent
+{
+    AnimationControlComponent* animationControl { nullptr };
+    // Keys are based on the animation index from the AnimationControl.
+    std::unordered_map<uint32_t, TransformAnimationSpline> animationSplines;
```
santiberna, line 20: Forgor to include unordered_map

santiberna (changes requested):
Addind on my previous review of moving the math wren bindings to another file, I think you should also move some of the animation logic to its .cpp instead of header. The linter also caught some missing include errors

engine/renderer/private/animation.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    if (!activeAnimation.has_value())
+    {
+        return std::nullopt;
+    }
+
+    return animations[activeAnimation.value()].name;
+}
+
+std::optional<uint32_t> AnimationControlComponent::CurrentAnimationIndex()
```
santiberna, line 82: Simply return the optional, no need  to check if it has a value

santiberna (approved):
Looks good on wren side. Approved!

PR #120, opened 2024-12-23
Fix normal transforms

MuskettaMan:
### Description

Normals transforms properly fixed. Normally this is implemented through a normal matrix using a transpose inverse. But this uses an adjoint matrix approach https://www.shadertoy.com/view/3s33zj. 

Screenshot_1
Before
Screenshot_2
After
image
Split

### Test criteria

- [ ] Check whether scaling objects now uses proper lighting calculations.

### DIFF
--- a/engine/renderer/private/renderer.cpp
+++ b/engine/renderer/private/renderer.cpp
@@ -93,7 +93,7 @@ Renderer::Renderer(ApplicationModule& application, Viewport& viewport, const std
 
     // temporary location
     auto font = LoadFromFile("assets/fonts/JosyWine-G33rg.ttf", 48, _context);
-    viewport.AddElement(std::make_unique<MainMenuCanvas>(_viewport.GetExtend(), _context, font));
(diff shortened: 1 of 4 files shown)


engine/renderer/private/renderer.cpp
```diff
@@ -93,7 +93,7 @@ Renderer::Renderer(ApplicationModule& application, Viewport& viewport, const std
 
     // temporary location
     auto font = LoadFromFile("assets/fonts/JosyWine-G33rg.ttf", 48, _context);
-    viewport.AddElement(std::make_unique<MainMenuCanvas>(_viewport.GetExtend(), _context, font));
+    //viewport.AddElement(std::make_unique<MainMenuCanvas>(_viewport.GetExtend(), _context, font));
```
dumitrurares14, line 96: remove commented code if not needed

dumitrurares14 (approved):
Looks good to me 🫡 Just remove the comment before merging otherwise all good

PR #121, opened 2025-01-06
Fix/perforce assets

santiberna:
### Description

This pull request moves all assets to Perforce and removes Git LFS.

### Test criteria

The reviewer should check if Perforce contains all the asset files and test run the game, AFTER mapping the Perforce Depot to the repo folder and getting latest.


mmzala (approved):
Nice, after configuring perforce it works. Maybe it is an idea to write a short github wiki page about configuring the perforce to get the assets correctly for the project? Would make it easier for others to set it up in case someone doesn't know how.

PR #122, opened 2025-01-06
Particles frustum culling + small bugfix

lyvanettinger:
### Description

- added frustum culling for particles
- fixed small bug with transformcomponent and emittercomponent interaction; when adding a transformcomponent to an entity with a rigidbody and spawn an emittercomponent, the particles kept spawning from (0, 0, 0) instead of the rigidbody location. This turned out to be because the physics system only updates the worldmatrix when there's also a staticmeshcomponent attached to the entity, which is not the case with my particle tests. This was (temporarily) fixed by getting the position directly from the rigidbody over getting it from the worldmatrix.

### Issues

https://bubonic-brotherhood.codecks.io/card/16o-frustum-culling-and-depth-collision

### Test criteria

- [ ] Create a plane, spawn emitters using 0 and watch the particle counter go down as you move your camera away from the particles
- [ ] Make sure the particles spawn from the debug drawn sphere


MuskettaMan (approved):
Looks good, just one minor code clean up comment

PR #123, opened 2025-01-06
Feature/frame graph reusable outputs

mmzala:
### Description

- Added reusable output resources to the fame graph

```cpp
FrameGraphNodeCreation uiPass { *_uiPipeline };
uiPass.SetName("UI pass")
    .SetDebugLabelColor(glm::vec3 { 255.0f, 255.0f, 255.0f })
    .AddOutput(_fxaaTarget, FrameGraphResourceType::eAttachment);

FrameGraphNodeCreation debugPass { *_debugPipeline };
debugPass.SetName("Debug pass")
    .SetDebugLabelColor(glm::vec3 { 0.0f, 1.0f, 1.0f })
    .AddInput(_gBuffers->Depth(), FrameGraphResourceType::eAttachment)
    .AddOutput(_fxaaTarget, FrameGraphResourceType::eAttachment);
```

Now you are just able to give the same output multiple times to the render passes and the data will accumulate to the same resource. When using, make sure the order of the passes is correct when giving it to the frame graph, as it can't automatically deduce the order in this case. There is also a new `allowSimultaneousWrites` optional argument for outputs, which allows the user to force the frame graph to not generate a memory barrier for this resource in the pass. It makes sure that the resources can be written to at the same time during multiple passes (e.g. lighting pass and sky dome pass).

- Converted the current passes to use reusable resources where possible
- Added a presentation pass to set up the swap chain image for presentation, as the frame graph doesn't support writing to the swap chain image currently. It just copies the last attachments data to the swap chain image.

### Issues

[Reusable outputs resources](https://bubonic-brotherhood.codecks.io/milestones/run/44/card/1f4-implement-reusable-output-resources)
[Converting render passes to use the reusable outputs](https://bubonic-brotherhood.codecks.io/milestones/run/44/card/1f5-convert-render-passes-to-use-reusable-output-resources)

### Test criteria

Visually, nothing should have changed. You can test with the following steps:
1. Start the program
2. The program should run as previously without any errors/warnings


MuskettaMan (approved):
Looks good to me. Tested and profiled and seem to work!

PR #125, opened 2025-01-06
Fix/audio blast

santiberna:
### Description

Fixes the initial audio blast by queueing sounds to only play after the fmod core system is updated (thus also updating all the 3D attributes). This might also be necessary to implement for events but my initial tests proved to work without the extra queuing.

### Issues

https://bubonic-brotherhood.codecks.io/decks/6-bugs/card/1hs-test-bgm-blast-for-first-frame

### Test criteria

Open the game and check if the background music is very high for the first frame


lyvanettinger (approved):
Looks all good and it's fixed on my end

PR #126, opened 2025-01-07
Pathfinding module

LarsJanssen21:
## Pathfinding module with initial algorithm draft
The pathfinding module with an initial first draft for a pathfinding algorithm

### Issues

https://bubonic-brotherhood.codecks.io/milestones/run/44/card/1ex-primitive-traversal-of-parsed-navigational-mesh

### Test criteria
- [ ] Run the engine in all configurations, it shouldn't crash and no new bugs should be introduced
- [ ] Ensure that on the perforce server this file is available \assets\models\NavmeshTest.gltf


LarsJanssen21:
Small update:
Files are now completely stored on perforce

engine/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-add_library(Engine INTERFACE)
+add_library(Engine INTERFACE
+        pathfinding/public/pathfinding_module.h)
```
MuskettaMan, line 2: this shouldnt have to be done manually, cmake can find the files it self, but i think you need to rename to hpp

MuskettaMan: but you shouldnt need to mention any specific headers in this file

engine/pathfinding/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+add_library(Pathfinding
+        private/pathfinding_module.cpp)
```
MuskettaMan, line 2: also shouldnt have to be done manually, refer to other module cmakelists to see how it is set up

engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -0,0 +1,258 @@
+#include "pathfinding_module.h"
+
+#include <engine.hpp>
```
MuskettaMan, line 3: make sure to use "" for includes that our from our code base. <> is used for third party includes

engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -0,0 +1,258 @@
...
+    ComputedPath path = this->FindPath(
+        { 0.7f, 0.0f, -0.7f },
+        { -0.7f, 0.0f, 0.7f });
+
+    return ModuleTickOrder::eTick;
+}
+
+void PathfindingModule::Tick(Engine& engine)
+{
+    (void)engine;
```
MuskettaMan, line 22: use MAYBE_UNUSED next to the argument, instead of this.

`MAYBE_UNUSED Engine& engine`

engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+void PathfindingModule::Shutdown(Engine& engine)
+{
+    (void)engine;
```
MuskettaMan, line 27: same here

engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+PathfindingModule::~PathfindingModule()
+{
+}
+
+int PathfindingModule::SetNavigationMesh(const std::string& mesh_path)
+{
+    CPUModel navmesh = _renderer->GetModelLoader().ExtractModelFromGltfFile(mesh_path);
+    CPUMesh navmesh_mesh = navmesh.meshes[0]; // GLTF model should consist of only a single mesh
```
MuskettaMan, line 41: make sure to use camelCase, so it's consistent with our codebase.

engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    // Loop over all triangles in the mesh
+    for (size_t i = 0; i < _triangles.size(); i++)
+    {
+        TriangleInfo& triangle = _triangles[i];
+
+        // We store all triangles that share indices with the current triangle
+        // If a triangle shares two indices with the current triangle, it's adjacent to the current triangle
+        std::multiset<uint32_t> shared_triangles;
```
MuskettaMan, line 78: sure you dont want to use an unordered_multiset?

engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -0,0 +1,258 @@
...
+    // G = totalCost
+    // H = estimateToGoal
+    // F = totalEstimatedCost
+
+    // Heuristic function
+    IterablePriorityQueue<TriangleNode, std::vector<TriangleNode>, std::greater<TriangleNode>> openList;
+    std::unordered_map<uint32_t, TriangleNode> closedList;
+
+    // Find closest triangle // TODO: More efficient way of finding closes triangle
+    float closestStartTriangleDistance = INFINITY, closestDestinationTriangleDistance = INFINITY;
```
MuskettaMan, line 146: would prefer to use numeric limits instead of defines, https://en.cppreference.com/w/cpp/types/numeric_limits

engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -0,0 +1,258 @@
...
+    const TriangleNode& finalTriangleNode = nodes[finalTriangleIndex];
+
+    pathNode.centre = _triangles[finalTriangleIndex].centre;
+    path.waypoints.push_back(pathNode);
+
+    uint32_t parentTriangleIndex = finalTriangleNode.parentTriangleIndex;
+
+    while (true)
+    {
+        if (parentTriangleIndex == UINT32_MAX)
```
MuskettaMan, line 247: https://en.cppreference.com/w/cpp/types/numeric_limits here too

engine/pathfinding/public/pathfinding_module.h
MuskettaMan, line 1: make sure to rename to hpp. can use `git mv old-name new-name` to keep history tracked

engine/pathfinding/public/pathfinding_module.h
```diff
@@ -0,0 +1,114 @@
...
+    std::string_view GetName() final { return "Pathfinding"; };
+
+public:
+    PathfindingModule();
+    ~PathfindingModule();
+
+    NON_COPYABLE(PathfindingModule)
+    NON_MOVABLE(PathfindingModule)
+
+    int SetNavigationMesh(const std::string& mesh);
```
MuskettaMan, line 94: use int32_t

MuskettaMan (changes requested):
Nice work, changes requrested are mostly related to code style, so should be simple to fix. I would have another look at the CMakeLists you changed, because that should be done differently, looking at other list files should make it clear.

engine/CMakeLists.txt
LarsJanssen21: Hmm weird, I don't recall adding this but it's fixed in the newest push!

engine/pathfinding/CMakeLists.txt
LarsJanssen21: Fixed in the newest push

engine/pathfinding/private/pathfinding_module.cpp
LarsJanssen21: Resolved in newest push

engine/pathfinding/private/pathfinding_module.cpp
LarsJanssen21: Resolved in newest push

engine/pathfinding/private/pathfinding_module.cpp
LarsJanssen21: Hadn't though about it, that does make more sense, updated in the newest push!

engine/pathfinding/private/pathfinding_module.cpp
LarsJanssen21: Fixed on all occasions

engine/pathfinding/public/pathfinding_module.h
LarsJanssen21: Fixed as well!

MuskettaMan (approved):

PR #127, opened 2025-01-07
Feature/particle wren bindings

lyvanettinger:
### Description

- Added wren bindings for particle module
- Added test with an emitter attached to the player

### Issues

https://bubonic-brotherhood.codecks.io/card/1e9-particle-interface-wren-bindings

### Test criteria

- [x] Take a look at the bitflags casting back and forth in particle_bindings.cpp
- [ ] Run the project and move around to see a trail of particles being left behind you


engine/bindings/private/particles/particle_bindings.cpp
```diff
@@ -0,0 +1,39 @@
...
+    self.SpawnEmitter(entity.entity, emitterPreset, static_cast<SpawnEmitterFlagBits>(flags), position, velocity);
+}
+}
+
+class ParticleFlagUtil
+{
+public:
+    static uint8_t GetSpawnEmitterFlagBits(SpawnEmitterFlagBits flag)
+    {
+        return static_cast<uint8_t>(flag);
```
santiberna, line 24: I think I also implemented some utility to reflect enums into wren. You should check that out. I use it for input bindings

game/game.wren
```diff
@@ -24,6 +24,9 @@ class Main {
 
             __player.AddAudioEmitterComponent()
 
+            var emitterFlags = ParticleFlagUtil.GetSpawnEmitterFlagBits(SpawnEmitterFlagBits.eIsActive()) | ParticleFlagUtil.GetSpawnEmitterFlagBits(SpawnEmitterFlagBits.eSetCustomVelocity())
```
santiberna, line 27: And I believe wren does support the bitflag operators, except you are limited to working with just numbers (wren does not have enums)

santiberna (changes requested):
The bindings seem fine, just revise a bit on the enum / bitflag situation. Wren does support enums and I did write a template for reflecting enums into wren, so please have a look at that.

Otherwise also update from main and setup your perforce

game/game.wren
lyvanettinger: It does, but how can I use those with the reflected enums in wren? Is it possible to cast the enums to uint in wren?

santiberna: Since enums dont natively exist in wren, what happens when they got bound it that their class is converted to a wren class and each enum value is converted to a static function that returns said enum value. The enum value returned is actually just a number in wren that supports all the bit operations

santiberna: To add: I believe wren only has one type for Numbers, which are doubles. The way it does bitshift operations with doubles is beyond me though, but the Wren documentation says that they are supported (maybe they are converted to integers under the hood). I think you should test this option, otherwise you will need to make a bitflag type for wren to work with.

santiberna (approved):
Seems correct now with the bitflag stuff! Merge it

PR #130, opened 2025-01-08
[UPDATE] Rename pipelines to passes

MuskettaMan:
### Description

Pipelines have been renamed to passes. Reasons for this:
1. The pipelines more often than not contain more than one pipeline. So it makes more sense to call it a pass, since they all work on the same resources, but have different commands working on it.
2. This frees up pipelines as a term in our codebase, so we can use that to further split up and abstract passes. This is important when dealing with the different pipelines that can be executed during the geometry pass, each requiring their own draw buffer.
3. This matches the naming with the frame graph.

### Test criteria

- [x] Still compiles and runs


mmzala (approved):
Looks good and it runs without any problems.

Just a heads up for splitting up passes that I just thought of. The frame graph does viewports, scissors and debug labels automatically for the pass. This would not happen if we would split up the pass into multiple steps that each have their own pipeline. It would only happen for the first pipeline in the pass. So either we would update the frame graph so it knows the pass has different steps or we don't do those automatically if we go that route.

PR #131, opened 2025-01-08
Feature/automatic steam build upload

mmzala:
### Description

- Adds a github action that automatically makes a build, packages it and sends it to Steam to set it active on the development branch. The build is triggered when a github release is made, or can be triggered manually through the actions tab
- Adds Steamcmd.exe back for uploading the builds

### Issues

[Codecks card](https://bubonic-brotherhood.codecks.io/milestones/run/51/card/1er-steam-upload-automation)

### Test criteria

There is not really a way to test this manually, as the github action script needs to be on the main branch for us to be able to manually start it.


.github/workflows/steam-build-upload.yml
```diff
@@ -0,0 +1,170 @@
...
+      
+      - name: Package Windows Project
+        if: matrix.platform == 'windows-latest'
+        run: |
+          python scripts/python/package.py -o "package/windows-dist" -e "build/x64-Release/CustomTech.exe" -a "assets" "shaders/bin" "game" -f "build/x64-Release/steam_api64.dll" "build/x64-Release/fmod.dll" "build/x64-Release/fmodstudio.dll"
+
+      - name: Package Linux Project
+        if: matrix.platform == 'ubuntu-latest'
+        run: |
+          python scripts/python/package.py -o "package/linux-dist" -e "build/WSL-Release/CustomTech" -a "assets" "shaders/bin" "game" -f "build/WSL-Release/libsteam_api.so" "build/WSL-Release/libfmod.so.14" "build/WSL-Release/libfmodstudio.so.14"
```
mmzala, line 99: I couldn't use the .bat scripts here, because on Linux you can't run them at all and on Windows it didn't run from the correct directory. I think we should make a .txt file where we put all the things that need to be included in the build, which both the .bat scripts and this automation can use.

MuskettaMan (approved):
Looks good! We can test it on main after merge

PR #132, opened 2025-01-08
Feature/UI controller support

luuk234:
Description

Adds controller support for UI navigation and input. 
input mode will switch to controller if a controller is available. currently it's not possible to switch input types during the game due to limitations with the input for now.

newly added logic is mostly in the different UI element update functions, and in the UIInputContext.

Issues

https://bubonic-brotherhood.codecks.io/milestones/run/44/card/1g7-ui-controller-support

testing procedure

1. in engine.cpp set the createNavigationTest to true
2. plug in controller or open on steamDeck.
3. check if one of the buttons is lit up (hovered) and if the right joystick allows you to switch buttons. if you press the trigger then the selected button should be clicked and a message should be logged.

performance has not been tested yet.


engine/source/engine.cpp
```diff
@@ -158,6 +160,15 @@ ModuleTickOrder OldEngine::Init(Engine& engine)
 
     applicationModule.GetActionManager().SetGameActions(GAME_ACTIONS);
 
+    // todo: move this to scripting.
```
VengarvdWal, line 163: Will this be changed in a later PR?

VengarvdWal (changes requested):
This PR makes it that when you load in the map without the UI active the camera movement is restricted, you can no longer turn 180 degrees or further. 

Also does it make sense that the UI requires the right joystick for the movement, and the right trigger to select instead of the SouthButton

engine/source/engine.cpp
luuk234: yes

lyvanettinger (commented):
Looks good and works fine on my end. Only point of critique is one that Vengar also brought up: to me it makes no sense that navigating the UI uses the right analog stick and right trigger. I'd personally expect something like left trigger or dpad for moving between buttons and either south button or east button for selection. 
Was there any reason why you went for right analog and trigger instead?

VengarvdWal (approved):
Looks good, tested it on the steamDeck aswell.

lyvanettinger (approved):
Looks all good!

PR #137, opened 2025-01-14
Feature/two pass hzb culling

MuskettaMan:
### Reviewer steps

### Description

Implementation of the two-pass gpu occlusion algorithm. This PR splits up the geometry and shadow drawing process in multiple passes. The idea behind the algorithm is to split up the geometry drawing in a first and second pass. The first draws what was visible in the previous frame and with that builds an HZB that is used to cull draws in the second pass. 

This also implements the use of the Count version of indirect draw. This reduces that amount of draws that have to be processed.

I've also added screenspace fog effect. 

Shadow buffers are moved to gpu scene from the gbuffers.

#### Profiling data

_Starting position_
Mesh rendering time on the Gpu (geom and shadow)
Before: 16ms
After: 13ms

_Close to the church_
Mesh rendering time on the Gpu (geom and shadow)
Before: 21ms
After: 12.5ms

### Issues

- [x] https://bubonic-brotherhood.codecks.io/card/1i1-finalize-gpu-occlusion-culling

### Test criteria

- [x] No popping of meshes is visible
- [x] The shadows are all drawn properly
- [x] Runs on the Steam Deck


LeoHellmig:
I am just taking a look and so far I have noticed that the scene looks way better. Better colors and nice fog.
image
I did see that the shadow for the skinned meshes does not look correct.
They don't seem to be in their animated state, but the gun does move when shooting.

image
This is what the shadows of skinned meshes look like on main.

engine/renderer/private/passes/generate_draws_pass.cpp
```diff
@@ -0,0 +1,181 @@
...
+        .srcQueueFamilyIndex = vk::QueueFamilyIgnored,
+        .dstQueueFamilyIndex = vk::QueueFamilyIgnored,
+        .buffer = buffers[0],
+        .offset = 0,
+        .size = sizeof(uint32_t),
+    };
+    fillBarriers[1] = fillBarriers[0];
+    fillBarriers[1].buffer = buffers[1];
+
+    commandBuffer.pipelineBarrier(vk::PipelineStageFlagBits::eTransfer, vk::PipelineStageFlagBits::eComputeShader, {}, {}, fillBarriers, {});
```
mmzala, line 63: A separate pass that can be reused for this, could be a good idea. Or maybe a new frame graph pass type that handles transfer operations? That way the barriers could again be handled automatically by the frame graph. Just suggestions for the future, I don't really mind if this stays like this for now.

shaders/generate_draws.comp
```diff
@@ -0,0 +1,306 @@
...
+            if (shouldDraw)
+            {
+                uint newIndex = atomicAdd(count, 1);
+                redirect[newIndex] = gId;
+                drawCommandsOut[newIndex] = drawCommandsIn[gId];
+                drawCommandsOut[newIndex].instanceCount = 1;
+            }
+
+            SetVisibility(gId, visible);
+        }
```
mmzala, line 108: Wouldn't is be cleaner to just split this up into 2 different shaders, instead having 1 shader do 2 things?

engine/renderer/private/passes/generate_draws_pass.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    if (_isPrepass)
```
mmzala, line 65: Wouldn't is be better to split it up into 2 different passes, instead of handling 2 things in 1 pass?

engine/utility/public/math_utils.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <cstdint>
+
+inline uint32_t DivideRoundingUp(
+    uint32_t _dividend,
+    uint32_t _divisor)
+{
+    return (_dividend + _divisor - 1) / _divisor;
```
mmzala, line 9: Why the `_` character for each argument name?

mmzala (commented):
Left a few comments, and it works great! it's possible I missed something, since it's a big PR. So it's good that a second person is assigned to it.

engine/renderer/private/camera_batch.cpp
```diff
@@ -0,0 +1,141 @@
...
+        .initialData = {},
+        .width = hzbSize,
+        .height = hzbSize,
+        .depth = 1,
+        .layers = 1,
+        .mips = static_cast<uint8_t>(std::log2(hzbSize)),
+        .flags = vk::ImageUsageFlagBits::eStorage | vk::ImageUsageFlagBits::eSampled,
+        .isHDR = false,
+        .format = vk::Format::eR16Sfloat,
+        .type = ImageType::eDepth, // TODO: should probably be color
```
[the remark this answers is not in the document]
MuskettaMan: It should be color, im not making use of the hardware depth testing, so color is fine to use. I think i also ran into issues before using depth

mmzala (approved):

luuk234 (approved):
behaves as expected and the code looks good as well, don't forget to remove the hdri.

PR #140, opened 2025-01-15
Feature/profiling simplify

MuskettaMan:
### Reviewer steps

### Description

- Update to tracy profiling zones, so it's a bit more organized. 
- Also makes sure systems in the ECS are automatically profiled. 
- Pipelines renamed to passes in tracy
- Cleaned up some unnecessary zones. Mainly cause their names are just profiled as "Update" making it hard to recognize what it actually is.

image
image
image
image

### Test criteria

- [x] Tracy results look organized


engine/ecs/private/ecs_module.cpp
```diff
@@ -44,6 +48,10 @@ void ECSModule::RenderSystems() const
     ZoneScoped;
     for (const auto& system : systems)
     {
+        ZoneScoped;
+        std::string name = std::string(system->GetName()) + " Render";
```
dumitrurares14, line 52: We should probably use const for these strings. No idea if it can help with anything but it's a good practice in my opinion. This std::string name is being done like this in multiple places, so if you decide to change it, it should be everywhere

dumitrurares14 (approved):
Beside that minor pick about const strings, everything looks good 💪

VengarvdWal (approved):
If you update the branch, it can be merged. Looks way nicer in Tracy

PR #153, opened 2025-02-11
Fix workflow deprecated action

santiberna:
This pull request fixes the workflow that uses a deprecated cache action version.

No review needed, PR-Checks should be sufficient.


mmzala (approved):

LarsJanssen21 (approved):
Wow! Stellar work!

PR #156, opened 2025-02-11
Fixed deprecated linter version

LarsJanssen21:
### Description

cpp-linter uses latest available release right now


santiberna (changes requested):
You should modify a .cpp file to really test if this fixes anything, since to run the linter only runs when code is changed.

LarsJanssen21:
> You should modify a .cpp file to really test if this fixes anything, since to run the linter only runs when code is changed.

Will do!

santiberna (approved):
Good job! Merging it

PR #158, opened 2025-02-13
Feature/gltf prefix parsing

luuk234:
### Reviewer steps

### Description

Adds in the ability to parse prefix objects from blender.
current prefixes that are checked but don't yet do anything are:
- PSPWN_ for the player spawn
- ESPWN_ for the enemy spawners

Also adds load flags for the modelloader to allow for only the chosen things to be loaded.
### Issues

https://bubonic-brotherhood.codecks.io/milestones/run/57/card/1o5-implementing-gltf-object-type-parsing

### Test criteria

1. In game_module.cpp uncomment the first model that is imported, titled: "prefix_test.glb", also comment the Cathedral.glb
2. get latest from perforce.
3. run the project
4. if asserts are thrown in the modelloader.cpp file for either prefix then The feature runs correctly.
5. also check if the other assets can this be loaded correctly.


engine/game/private/scene/scene_loader.cpp
```diff
@@ -23,6 +23,23 @@
...
+
+    // player spawn point
+    // note: if there are multiple spawnpoints in the given hierarchy then the last one to be processed will be the actual player spawn point.
+    if (node.name.starts_with("PSPWN_"))
+    {
+        assert(false && "Setting player spawn not yet implemented");
+    }
+}
+
 void LoadNodeRecursive(ECSModule& ecs,
```
santiberna, line 43: Improvement point for API: in order to make this function signature a lot cleaner, it is possible to refactor the recursive node initialization into a class `RecursiveNodeLoader` that is initialized with most stuff that stays consistent throughout the loading process: hierarchy, model, cpu model, entity LUT, animation control... Then all we need is to call an Instantiate function that just accepts a parent, node index... (all the step specific parameters)

engine/game/private/scene/scene_loader.cpp
```diff
@@ -58,52 +76,66 @@ void LoadNodeRecursive(ECSModule& ecs,
             ecs.GetRegistry().emplace<StaticMeshComponent>(entity).mesh = model.staticMeshes.at(currentNode.meshIndex.value().second);
 
             // check if it should have collider
-
-            ecs.GetRegistry().emplace<RigidbodyComponent>(entity, ecs.GetSystem<PhysicsSystem>()->CreateMeshColliderBody(cpuModel.meshes.at(currentNode.meshIndex.value().second), PhysicsShapes::eCONVEXHULL, entity));
-
-            // add collider recursively
-
+            if (HasAnyFlags(loadFlags, SceneLoading::LoadFlags::eLoadCollision))
```
santiberna, line 79: Missing an include for HasAnyFlags

engine/game/private/scene/scene_loader.cpp
```diff
@@ -58,52 +76,66 @@ void LoadNodeRecursive(ECSModule& ecs,
...
     for (const auto& nodeIndex : currentNode.children)
     {
         const entt::entity childEntity = ecs.GetRegistry().create();
-        LoadNodeRecursive(ecs, childEntity, nodeIndex, hierarchy, entity, model, cpuModel, animationControl, entityLUT, skeletonRoot);
+        LoadNodeRecursive(ecs, childEntity, nodeIndex, hierarchy, entity, model, cpuModel, animationControl, entityLUT, skeletonRoot, false, loadFlags);
     }
 }
 
-entt::entity SceneLoading::LoadModelIntoECSAsHierarchy(ECSModule& ecs, const GPUModel& gpuModel, const CPUModel& cpuModel, const Hierarchy& hierarchy, std::vector<Animation> animations)
+entt::entity SceneLoading::LoadModelIntoECSAsHierarchy(ECSModule& ecs, const GPUModel& gpuModel, const CPUModel& cpuModel, const Hierarchy& hierarchy, std::vector<Animation> animations, LoadFlags loadFlags)
```
santiberna, line 138: pass animations as const ref

engine/game/private/scene/scene_loader.cpp
```diff
@@ -120,6 +152,8 @@ entt::entity SceneLoading::LoadModelIntoECSAsHierarchy(ECSModule& ecs, const GPU
     if (hierarchy.skeletonRoot.has_value())
     {
         entt::entity skeletonEntity = ecs.GetRegistry().create();
+
+        // Note: load twice?
         LoadNodeRecursive(ecs, skeletonEntity, hierarchy.skeletonRoot.value(), hierarchy, entt::null, gpuModel, cpuModel, animationControl, entityLUT, entt::null, true);
```
santiberna: This seems to be a mistake or a merge issue

engine/game/public/scene/scene_loader.hpp
```diff
@@ -14,5 +14,18 @@ struct Animation;
...
+{
+    eLoadMeshes = 1 << 0,
+    eLoadSkeletalMeshes = 1 << 1,
+    eLoadCollision = 1 << 2,
+    eLoadGameplayComponents = 1 << 3,
+    eLoadLights = 1 << 4,
+    eAll = eLoadMeshes | eLoadSkeletalMeshes | eLoadCollision | eLoadGameplayComponents | eLoadLights
 };
+
+GENERATE_ENUM_FLAG_OPERATORS(LoadFlags);
```
santiberna, line 26: missing the include for this macro

santiberna (changes requested):
Short PR so not much to say, I gave one API improvement suggestion that would clean up a lot of the scene loader logic but check if its actually worth. Another concern is that there should probably be a documented and easy to access place where all the special GLTF naming and setup conventions that we have are described.

engine/game/private/scene/scene_loader.cpp
luuk234: since ferri is also going to be looking at the animation and will touch this function I will keep the signature as is

engine/game/private/scene/scene_loader.cpp
luuk234: this works currently, so i will not touch it, already communicated with Ferri about this

luuk234:
also I will write up a wiki article for the prefixes

santiberna (approved):
Wiki article was created and the rest of the PR looks good! Accepted

engine/game/private/scene/scene_loader.cpp
```diff
@@ -23,6 +23,23 @@
...
+    // enemy spawners
+    if (node.name.starts_with("ESPWN_"))
+    {
+        assert(false && "Enemy spawner component not yet implemented");
+        // ecs.GetRegistry().emplace<EnemySpawnerComponent>(currentEntity);
+    }
+
+    // player spawn point
+    // note: if there are multiple spawnpoints in the given hierarchy then the last one to be processed will be the actual player spawn point.
+    if (node.name.starts_with("PSPWN_"))
```
santiberna, line 37: Why do we not use a camera for our player start?

santiberna (changes requested):
PR is quite old now and there is no direct way to test it yet... maybe we should delay this further until we have the blockout level

MuskettaMan (commented):
Mainly looks good, but hard to tell since it isn't used yet. The part for string checking on the node names looks good.

Also notice a lot of changes to the scene loader, and you need to update from main, I think that could be a hard merge to implement. That also makes it hard right now to review the changes to the sceneloader since it is outdated.

luuk234:
sceneloader changed too much, will redo this pr

PR #168, opened 2025-02-19
Feature/warnings as errors

mmzala:
### Reviewer steps

### Description

Enabled warnings as errors. Also fixes the warnings we already had.
The warnings are already considered as errors in our CI/CD, so I didn't have to change anything there (which somehow didn't detect 2 warnings we had).

### Issues

[Codecks card](https://bubonic-brotherhood.codecks.io/milestones/run/57/card/1o4-warnings-as-errors)

### Test criteria

- [ ] The project should compile successfully

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ include(scripts/cmake/code-modules.cmake)
 
 ### COMPILATION SETTINGS ###
 
-option(WARNINGS_AS_ERRORS "Enable warnings as errors" OFF)
+option(WARNINGS_AS_ERRORS "Enable warnings as errors" ON)
 
 option(COMPILE_TESTS "Compile with Unit Tests" ON)
 option(COMPILE_SHADERS "Compile all GLSL shaders as part of build step" ON)
--- a/engine/physics/private/systems/physics_system.cpp
+++ b/engine/physics/private/systems/physics_system.cpp
@@ -32,8 +32,7 @@ entt::entity PhysicsSystem::LoadNodeRecursive(const CPUModel& models, ECSModule&
     const Hierarchy& hierarchy,
     entt::entity parent, PhysicsShapes shape)
 {
-
-    if (const bool validation = shape != eMESH && shape != eCONVEXHULL)
+    if (MAYBE_UNUSED const bool validation = shape != eMESH && shape != eCONVEXHULL)
     {
         assert(!validation && "Shape is not supported, please use eMESH or eCONVEXHULL");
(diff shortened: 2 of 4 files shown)


santiberna (changes requested):
This pull request really does nothing, because warnings are enabled as errors in our build checks already. What may be happening is that the Github runners have a different version of clang has less warnings than the version we use? Also note that we do not care about warnings in our dependencies.

Playing around with the PR also showed that warnings as errors are defined for Release and Debug configs which is not useful for development.

mmzala:
> This pull request really does nothing, because warnings are enabled as errors in our build checks already. What may be happening is that the Github runners have a different version of clang has less warnings than the version we use? Also note that we do not care about warnings in our dependencies.
> 
> Playing around with the PR also showed that warnings as errors are defined for Release and Debug configs which is not useful for development.

As a team during our last retro of Block B to enable warnings as errors in our development environment as well. This is because of CI/CD already had them, which made our configs in different places not consistent. And there is no reason we should allow warnings to be there in the first place.

I don't really understand you last point as well. How is it not useful? Why should be allow warnings?

santiberna:
Oh okay I did not understand that it was agreed with the team that warnings need to be enabled as errors even in development debug / release builds. This does kind of defeat the whole purpose of the CI/CD config. I thought there was an issue with PR Checks and they were not detecting warnings as errors. My apologies.

I my opinion however, warnings as errors is really good for a codebase to follow, but very annoying when actually working on and prototyping code. 

If we do go forward with this, then you could just delete the CI/CD config. Are you sure that the entire team still agrees on this though?

mmzala:
> Oh okay I did not understand that it was agreed with the team that warnings need to be enabled as errors even in development debug / release builds. This does kind of defeat the whole purpose of the CI/CD config. I thought there was an issue with PR Checks and they were not detecting warnings as errors. My apologies.
> 
> I my opinion however, warnings as errors is really good for a codebase to follow, but very annoying when actually working on and prototyping code.
> 
> If we do go forward with this, then you could just delete the CI/CD config. Are you sure that the entire team still agrees on this though?

It was one of the action points from the retro that is still written down on the miro board, so I'm quite sure that everyone besides you saw this. If you want we can talk to the team tomorrow and check if they still want to do it, I think that would be best to do.

Also we can't delete the CI/CD config, otherwise there will be no CI/CD at all, since the warnings as errors settings is set as a parameter for cmake.

mmzala:
The team decided to not do this after all. Closing PR.

PR #172, opened 2025-02-24
Feature/animation hierarchy optimization

MuskettaMan:
### Reviewer steps

### Description

Implemented update to optimize the calculation of world transforms for animated models. Skeletons now have their own separate hierarchy that's updated once. This removes duplicate updates, and allows for a top-down traversal approach.

image
image

Before  5.77ms
After 1.45ms

This also adds a few improvements to the animation system and makes sure everything runs correctly.

### Test criteria

 - [ ] Check my architecture for the skeleton hierarchy, at times this feels a bit shoe horned in. Perhaps this can be separated better?
 - [ ] Check whether the animations run smoothly
 - [ ] Tell me to change back to the cathedral before merging

### DIFF
--- a/.github/workflows/steam-build-upload.yml
+++ b/.github/workflows/steam-build-upload.yml
@@ -91,12 +91,12 @@ jobs:
       - name: Package Windows Project
         if: matrix.platform == 'windows-latest'
         run: |
-          python scripts/python/package.py -o "package/windows-dist" -e "build/x64-Release/CustomTech.exe" -a "assets" "shaders/bin" "game" -f "settings.json" "build/x64-Release/steam_api64.dll" "build/x64-Release/fmod.dll" "build/x64-Release/fmodstudio.dll"
+          python scripts/python/package.py -c "scripts/build_system/package_config/windows_dist.json"
 
       - name: Package Linux Project
         if: matrix.platform == 'ubuntu-latest'
         run: |
-          python scripts/python/package.py -o "package/linux-dist" -e "build/WSL-Release/CustomTech" -a "assets" "shaders/bin" "game" -f "settings.json" "build/WSL-Release/libsteam_api.so" "build/WSL-Release/libfmod.so.14" "build/WSL-Release/libfmodstudio.so.14"
+          python scripts/python/package.py -c "scripts/build_system/package_config/linux_dist.json"
 
       # Save build as artifact for upload job
 
--- a/scripts/build_system/package_and_upload_steam_dev.bat
+++ b/scripts/build_system/package_and_upload_steam_dev.bat
@@ -1,6 +1,6 @@
 @echo off
-call package_dist_linux.bat
-call package_dist_windows.bat
+call package_linux_dist.bat
+call package_windows_dist.bat
 echo Uploading to the development branch.
 set /p username="Steam username: "
 set /p password="Steam password: "
--- a/scripts/build_system/package_and_upload_steam_public.bat
+++ b/scripts/build_system/package_and_upload_steam_public.bat
@@ -1,6 +1,6 @@
 @echo off
-call package_dist_linux.bat
-call package_dist_windows.bat
+call package_linux_dist.bat
+call package_windows_dist.bat
 echo Uploading to the public branch. Setting the build live is not done automatically for public builds, that must be done via the admin panel!
 set /p username="Steam username: "
 set /p password="Steam password: "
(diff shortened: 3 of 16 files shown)


santiberna (changes requested):
In general, I already suggested a couple improvement points but I will record them here:

- I feel like you can simplify the entire component data layouts by separating all immutable data related to skeletons and animations into a dedicated Skeleton and AnimationSet resource types that can be owned by multiple entities (gives us flyweight if we have multiple similar models with the same animations). 
- Work on correctly naming each component -> Skeleton for data related to the main model entity, Joint for each node in a skeleton, Animation for anything that manages the current animation state
- Add tracy zones to check impact of each stage of updating the joint transforms and sending the data to the GPU

The resource types can be delegated to another pull request, that you should do after this. Otherwise the implementation does seem correct. Also find correct animations to test out your PR since there can be still some weird issue hidden under the buggy dinosaurs

engine/renderer/public/components/skeleton_component.hpp
```diff
@@ -1,8 +1,34 @@
...
+{
+    entt::entity parent;
+    std::array<entt::entity, 7> children; // TODO: Check optimal size.
+};
 
 struct SkeletonComponent
 {
-    uint32_t boneOffset = 0;
+    entt::entity root;
+};
```
LeoHellmig, line 16: default initialize to entt::null

engine/renderer/private/components/skeleton_component.cpp
```diff
@@ -0,0 +1,24 @@
...
+}
+
+void SkeletonHelpers::InitializeSkeletonNode(SkeletonNodeComponent& node)
+{
+    node.parent = entt::null;
+    for (auto& child : node.children)
+    {
+        child = entt::null;
+    }
+}
```
LeoHellmig, line 24: Could be removed by default initializing the component to entt::null

engine/game/private/scene/scene_loader.cpp
```diff
@@ -14,98 +14,165 @@
...
 
-    for (const auto& nodeIndex : currentNode.children)
+private:
+    ECSModule& _ecs;
+    const Hierarchy& _hierarchy;
+    AnimationControlComponent* _animationControl;
+
+    SkeletonComponent* _skeletonComponent;
+
+    void LoadNode(entt::entity entity, uint32_t currentNodeIndex, entt::entity parent)
```
LeoHellmig, line 136: Can be named RecursiveLoadNode

engine/inspector/private/editor.cpp
```diff
@@ -100,7 +100,7 @@ void Editor::DrawHierarchy()
             {
                 RelationshipComponent* relationship = _ecs.GetRegistry().try_get<RelationshipComponent>(entity);
 
-                if (relationship == nullptr || relationship->parent == entt::null)
+                if (relationship != nullptr && relationship->parent == entt::null)
                 {
```
LeoHellmig, line 104: This should not be changed
It will make orphans not show up in the world hierarchy

LeoHellmig (changes requested):
I found a few nitpicks and changes that should be made, but the feature works well, nice

engine/renderer/private/components/skeleton_component.cpp
MuskettaMan: Sadly can't, since the children in the array are initialized to entity 0, instead of null

engine/ecs/public/components/relationship_component.hpp
```diff
@@ -14,6 +14,10 @@ struct RelationshipComponent
     entt::entity parent = entt::null;
 };
 
+struct HideOrphan
```
santiberna, line 17: What is the purpose of this tag?

santiberna (approved):
Well everything else seems correct (and the animation work as demonstrated with the new animated monster). I have just one question that you can answer in the comments. 

Otherwise I think you can merge and start considering about the possible improvements we talked about (if a priority)

engine/ecs/public/components/relationship_component.hpp
MuskettaMan: Something I discussed with Leo, for most entities we want to see if theyre orphans so we can still work with them, or if there are entities leaking. But in the case of the joints we dont want to see them cause it bloats. So the HideOrphan can be applied to entities that we dont want to see in the editor

LeoHellmig (approved):
Looks good, I tested and the hierarchy shows the correct orphans. Well done

PR #182, opened 2025-03-10
Feature/particle sprite sheets

lyvanettinger:
### Reviewer steps

### Description

https://github.com/user-attachments/assets/7cec2dd0-db80-4ffb-a79d-657289f19121

- Added option to load sprite sheets for particle sprite
- Added frame blending for sprite sheets
- Added variables to ImGui menus to edit for particle sprite sheets

### Issues

https://bubonic-brotherhood.codecks.io/milestones/run/58/card/1jf-particle-spritesheets 

### Test criteria

!! Pull from Perforce before running !!
- [x] There should be a flame sprite sheet spawned on the little island left from the cathedral
- [x] Spawn a new particle with sprite sheet (in particle editor, edit spritesheet test with new sprite sheet, edit sprite dimension and frame count to match the sprite sheet and toggle frameblend on/off. make sure to adjust the new particle's velocity in the hierarchy editor to 0,0,0 so they don't fly off first thing)


engine/particles/private/particle_module.cpp
```diff
@@ -86,50 +86,57 @@ void ParticleModule::Tick(MAYBE_UNUSED Engine& engine)
...
 
-void ParticleModule::SetEmitterPresetImage(EmitterPreset& preset, ResourceHandle<GPUImage> image)
+bool ParticleModule::SetEmitterPresetImage(EmitterPreset& preset, std::string fileName)
 {
     auto resources = _context->Resources();
 
+    bool imageFound;
+    auto image = GetEmitterImage(fileName, imageFound);
+
+    preset.imageName = fileName;
```
luuk234, line 125: std:::move from string

engine/particles/private/particle_module.cpp
```diff
@@ -140,14 +147,12 @@ void ParticleModule::LoadEmitterPresets()
         preset.flags = static_cast<uint32_t>(ParticleRenderFlagBits::eNoShadow);
         preset.color = glm::vec4(1.0f);
         preset.name = "Test";
-        SetEmitterPresetImage(preset, image);
+        SetEmitterPresetImage(preset, "jeremi.png");
```
luuk234, line 150: fine for now, likely want to parameterize this

shaders/particles/billboard.vert
```diff
@@ -20,29 +20,56 @@ layout (location = 3) in vec2 inTexCoord;
...
+    uv2 += instance.frameOffsetNext;
+    uv2 *= instance.textureMultiplier;
+    texCoord2 = uv2;
+
     normal = normalize((camera.view * vec4(inNormal, 0.0)).xyz);
     materialIndex = instance.materialIndex;
-    texCoord = inTexCoord;
     flags = instance.flags;
     outColor = instance.color;
+    frameBlend = instance.frameBlend;
```
luuk234, line 72: if performance of this is fine then it's good. Seems like a lot of instructions for a simple billboard tho

luuk234 (approved):
overall looks good, keep the comments in mind but it's not important enough to delay this pr.

lyvanettinger:
As reply to Santiago's feedback:

> * The load successful message appears in the editor even if the image loading failed
I've changed it up slightly to give an "Error when loading image" message in editor.

> * The file path for the emitter presets should be already cached (I shouldn't need to reinput the path when opening the engine)
Implemented now.

> * Cant edit the initial velocity attribute in the preset menu, but I can in the entity editor (consider using the same struct for both the preset and component)
Also implemented now :)

> * Does sprite dimensions refer to the number of sprites in x and y axis or the size of each sprite? (you can also manually specify frame count after this, which is unintuitive because if sprite dimensions refers to the former, then this attribute is a little useless imo)
> 
> A couple issues as well:
> 
> * The flame sprite seems to vanish for one frame (the spritesheet has an empty sprite at the end)
So regarding this issue and the editor window being separate for amount of frames and x and y dimensions, it's to prevent this issue from happening, so I'm not sure why it's happening (it shouldn't). It might be that the particle lifetime is shorter than the frames shown/emit delay so it disappears before the next one's spawn.

> * The rotation towards the player seems to be incorrect: these two screenshots show the player in the same position, but looking at different angles.
I've made a bug report for this issue as I wasn't able to solve this at the moment: https://bubonic-brotherhood.codecks.io/decks/6-qa/card/1qu-particle-billboards-rotate-weird

shaders/particles/billboard.vert
lyvanettinger: I agree that the particle struct is getting quite bloated, but I'm not entirely sure how to circumvent this. Possibly I could cache some common variables over certain particle batches from the same emitter in an emitter struct they reference, but that would require some looking into. I can throw it on the backlog to look at later because I do feel like with the amount of particles spawned, it is eating up quite some memory

santiberna (approved):
Feedback was implemented and the bug ticket was created! Good job, everything else looks fine

PR #185, opened 2025-03-10
Primitive enemy traversing terrain

LarsJanssen21:
### Reviewer steps

### Description

A primitive enemy moves through the environment making use of the pathfinding and physics modules
EnemyFollowingPath

### Test criteria

The wren script should run without any errors. The enemy spawns at the village bridge and moves into the cathedral stopping there when the altar has been reached. Reloading the script should remove the existing enemy and spawn in a new one at the bridge


game/gameplay/enemies/enemies.wren
```diff
@@ -0,0 +1,93 @@
+import "engine_api.wren" for Vec3, Engine, ShapeFactory, Rigidbody, RigidbodyComponent, CollisionShape, Math
+
+class MeleeEnemy {
+
+    construct new(spawnPosition, engine) {
+        if(__engine == null) {
```
santiberna, line 6: Do not store the engine as a static var, prefer to pass it in the update function as well

game/gameplay/enemies/enemies.wren
```diff
@@ -0,0 +1,93 @@
...
+            __engine = engine
+        }
+
+        if(__physicsShape == null) {
+            __physicsShape = ShapeFactory.MakeCapsuleShape(1.7, 0.5)
```
santiberna, line 11: Physics shape could be passed from constructor

game/gameplay/enemies/enemies.wren
```diff
@@ -0,0 +1,93 @@
...
+        var startPos = body.GetPosition()
+        _currentPath = __engine.GetPathfinding().FindPath(startPos, Vec3.new(-16.0, 29.0, 195.1))
+        _currentPathNodeIdx = 0
+
+        var waypoint = _currentPath.GetWaypoints()[_currentPathNodeIdx]
+        var velocity = (waypoint.center - startPos).normalize() * Vec3.new(__maxVelocity, __maxVelocity, __maxVelocity)
+        body.SetVelocity(velocity)
+    }
+
+    Create(spawnPosition) {
```
santiberna, line 73: This function should be in construct new()

game/gameplay/enemies/spawner.wren
```diff
@@ -0,0 +1,45 @@
...
+
+    ClearAllEnemies() {
+        for(enemy in _meleeEnemies) {
+            enemy.Destroy()
+        }
+    }
+
+    Destroy() {
+        this.ClearAllEnemies()
+        __engine = null
```
santiberna, line 42: This is an error, calling destroy on a spawner will also set the __engine of all other spawners to null

santiberna (changes requested):
Great work on getting the enemy moving on the path with physics! However, I can see some small issues with your wren implementation and also on some of the work you did for the renderer bindings:
- Try to avoid passing and storing the engine as a static var for wren objects, prefer passing it in the functions that need it
- We do not need renderer bindings for loading CPUModels into GPUModels, or adding Animation Control Components, we simply need a function that accepts a path and returns an entity which is the root of the model that was loaded.
- The enemy rotating towards the destination seems to be a bit bugged? maybe stick to only rotating around the Y axis.
- Get the spawner to spawn every X amount of seconds (I want to see more than one enemy).

Regardless, if you fix the above stuff and make a nice video it would be amazing for the presentation!

engine/bindings/private/game/game_bindings.cpp
```diff
@@ -3,13 +3,17 @@
...
 #include "components/transform_component.hpp"
 #include "components/transform_helpers.hpp"
+#include "cpu_resources.hpp"
 #include "ecs_module.hpp"
 #include "entity/wren_entity.hpp"
 #include "game_module.hpp"
+#include "model_loading.hpp"
 #include "physics/shape_factory.hpp"
 #include "physics_module.hpp"
+#include "renderer_module.hpp"
```
LeoHellmig, line 16: Do these includes new need to be here?

LeoHellmig (approved):
Aside from Santiago' comments I have only found some redundant includes.

PR #193, opened 2025-03-18
added resolution scaling

luuk234:
### Reviewer steps

### Description

this feature adds scaling of the UI based on the screen resolution.
It can also be set manually for accesability options. 

### Issues

https://bubonic-brotherhood.codecks.io/milestones/run/64/card/1oy-fix-the-resolution-scaling-of-the-ui

### Test criteria

- [x] set the resolution of your monitor to different sizes and check if the UI still displays correctly.
- [ ] check if it looks correct on the steamdeck.


LeoHellmig (approved):
Looks good

engine/game/private/ui/ui_menus.cpp
```diff
@@ -74,7 +74,7 @@ std::pair<std::unique_ptr<Canvas>, HUD> HudCreate(GraphicsContext& graphicsConte
     std::unique_ptr<Canvas> canvas = std::make_unique<Canvas>(screenResolution);
 
     // temporary
-    canvas->SetAbsoluteTransform(canvas->GetAbsoluteLocation(), canvas->GetRelativeScale());
+    canvas->SetAbsoluteTransform(canvas->GetAbsoluteLocation(), screenResolution);
```
santiberna, line 77: Why is this temporary?

engine/user_interface/private/ui_module.cpp
```diff
@@ -8,7 +8,9 @@ ModuleTickOrder UIModule::Init(Engine& engine)
 {
     const glm::vec2 extend = engine.GetModule<ApplicationModule>().DisplaySize();
     _viewport = std::make_unique<Viewport>(extend);
-    _graphicsContext = engine.GetModule<RendererModule>().GetGraphicsContext();
+    Viewport::SetScaleMultiplier(Viewport::CalculateScaleMultiplierFromVerticalResolution(extend.y));
+    _graphicsContext
+        = engine.GetModule<RendererModule>().GetGraphicsContext();
```
santiberna, line 13: remove random newline

engine/user_interface/private/viewport.cpp
```diff
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <ranges>
 
+float Viewport::_uiScaleMultiplier = 0.5f;
```
santiberna, line 6: should be constexpr or class member to viewport

santiberna (changes requested):
Code in general looks good, but I have a couple questions regarding the implementation:
- Has it been tested on Steamdeck?
- What is the scale for relative positioning of the objects?

Other mini questions in specific spots. Respond to them plz plz

engine/user_interface/private/viewport.cpp
luuk234: because this is much easier, otherwise you have to keep track of this value for every uielement which is a pain, it is very unlikely that we will ever need more than 1 multiplier so making it static is a lot easier

luuk234: i don't think it's a problem

luuk234: can't be constexpr also because it's set during runtime

luuk234:
you can test it by manually setting the resolution of your monitor, it has not been tested on the steamdeck

engine/user_interface/private/viewport.cpp
santiberna: > because this is much easier, otherwise you have to keep track of this value for every uielement which is a pain, it is very unlikely that we will ever need more than 1 multiplier so making it static is a lot easier

Why is it not a class member to viewport then? (unless viewport is a namespace, which also doesn't sound right)

santiberna:
> arity

santiberna:
> you can test it by manually setting the resolution of your monitor, it has not been tested on the steamdeck. scale is currently set to 1440p as this was the res that was used while making the UI. when we remake the HUD this should be set to 1080p or 720p for clarity

image
image

Pictures taken at different resolutions, so it works!

santiberna (approved):

PR #195, opened 2025-03-18
CPUModel caching and physics colliders

santiberna:
### Reviewer steps

### Description

This PR adds two major caching features: the cpu model now stores generated collision, which means that it is not generated every model instantiation (could be later expanded with name prefixes) and adds a SceneLoader class instead of the SceneLoading namespace, that handles the caching of both CPU and GPU model data.

This was also updated in the scene loading wren bindings and it means that we can now hotreload much faster and not hit the batch buffer cap

### Issues

https://bubonic-brotherhood.codecks.io/card/1qy-cache-loaded-models
https://bubonic-brotherhood.codecks.io/card/1r1-move-jolt-collision-generation-to-cpumodel

### Test criteria

- [x] Check if hot reloading is faster
- [x] Check if the batch buffer data limit is not hit after hot reloading a bunch of times


engine/renderer/public/cpu_resources.hpp
```diff
@@ -10,109 +10,9 @@
...
-    std::vector<CPUMaterial> materials {};
-    std::vector<CPUImage> textures {};
-
-    std::vector<Animation> animations {};
-
-    std::string name {};
-};
+#include <resources/hierarchy.hpp>
+#include <resources/mesh.hpp>
+#include <resources/model.hpp>
```
LeoHellmig, line 15: use "" instead of <>

engine/model_loading/public/resources/hierarchy.hpp
```diff
@@ -0,0 +1,61 @@
+#pragma once
+#include <glm/mat4x4.hpp>
+#include <glm/vec3.hpp>
+#include <resources/mesh.hpp>
+
```
LeoHellmig, line 5: use ""

LeoHellmig (changes requested):

engine/bindings/private/wren_bindings.cpp
```diff
@@ -39,7 +39,10 @@ void TransitionToScript(WrenEngine& engine, const std::string& path)
 
 WrenEntity LoadModelScripting(WrenEngine& engine, const std::string& path)
 {
-    auto entity = SceneLoading::LoadModel(*engine.instance, path);
+    auto& sceneCache = engine.instance->GetModule<GameModule>()._scenesLoaded;
```
luuk234, line 42: are scene and model used interchangably here?

santiberna: Yeah I personally find the entire naming a bit confusing aswell, should I change it?

LeoHellmig (approved):
looks good!

PR #197, opened 2025-03-19
Movement update

dumitrurares14:
### Reviewer steps

### Description

https://github.com/user-attachments/assets/21e46455-1264-49c2-9bf2-276f4668a69b

This PR adds some solid changes to the movement. 

- Added a new dash system that works with ray casts instead of applying direct velocities (This was made in order to get rid of the gravity feel while dashing, making it more snappy)
- Added interpolation for slide direction to make sudden turns durns or changes in movement input less abrupt while sliding
- Added interpolation between crouch and stand height to fix the camera snap effect when changing between sliding and the rest of the movement

### Issues

https://bubonic-brotherhood.codecks.io/card/1s3-tweak-movement

### Test criteria

- [ ] Play around in the level and let me know if the changes improve the feel of the game
- [ ] See if you can break it 
- [ ] Let me know if anything behaves differently than what you would expect.


engine/bindings/private/physics/physics_bindings.cpp
```diff
@@ -50,6 +50,11 @@ void SetFriction(WrenComponent<RigidbodyComponent>& self, const float friction)
     self.component->SetFriction(friction);
 }
 
+void SetTranslation(WrenComponent<RigidbodyComponent>& self, const glm::vec3& translation)
```
santiberna, line 53: This function already exists for Transform components, why does it need to be defined for Rigidbodies? Should be handled more gracefully

engine/bindings/private/physics/physics_bindings.cpp
```diff
@@ -85,6 +90,11 @@ glm::vec3 GetRayHitNormal(RayHitInfo& self)
     return self.normal;
 }
 
+float GetRayHitFraction(RayHitInfo& self)
```
santiberna, line 93: These bindings could actually be directly implemented in the RayHitInfo class... just a suggestion

engine/bindings/private/utility/math_bind.hpp
```diff
@@ -201,6 +205,7 @@ inline void BindMath(wren::ForeignModule& module)
         mathUtilClass.funcStatic<&MathUtil::ToDirectionVector>("ToVector");
         mathUtilClass.funcStatic<&MathUtil::ToQuat>("ToQuat");
         mathUtilClass.funcStatic<&MathUtil::Mix>("Mix");
```
santiberna, line 207: Maybe rename to MixVec3? To keep consistent with naming

engine/physics/public/components/rigidbody_component.hpp
```diff
@@ -32,6 +32,7 @@ class RigidbodyComponent
     void SetAngularVelocity(const glm::vec3& velocity) { bodyInterface->SetAngularVelocity(bodyID, ToJoltVec3(velocity)); };
     void SetGravityFactor(float factor) { bodyInterface->SetGravityFactor(bodyID, factor); }
     void SetFriction(float friction) { bodyInterface->SetFriction(bodyID, friction); }
+    void SetTranslation(const glm::vec3& translation) { bodyInterface->SetPosition(bodyID, ToJoltVec3(translation), JPH::EActivation::Activate); }
```
santiberna, line 35: In what situation do we need to manually set the translation of rigidbody components in script?

game/gameplay/movement.wren
```diff
@@ -186,7 +194,7 @@ class PlayerMovement{
         var isJumpHeld = engine.GetInput().GetDigitalAction("Jump").IsHeld()
         var doubleJump = engine.GetInput().GetDigitalAction("Jump").IsPressed()
 
-        if(isGrounded && isJumpHeld) {
+        if((isGrounded && isJumpHeld)) {
```
santiberna, line 197: extra parentheses for no reason

game/gameplay/movement.wren
```diff
@@ -215,14 +222,18 @@ class PlayerMovement{
...
             }
 
+            //add slide here
+            if(slideWishDirection.length() > 0.01){
+                velocity = velocity + slideWishDirection.mulScalar(slideForce)
+            }
+
             var speed = velocity.length()
-            if (speed > maxSpeed) {
+            if (speed > maxSpeed ) {
```
santiberna, line 231: space for no reason

santiberna (changes requested):
In general the implementation makes movement a lot more cohesive and less buggy, even if there are still some issues with collision. The code looks alright to me, even if there are some formatting issues and questions that I highlighted in the comments! Just answer them and I will approve!

lyvanettinger (approved):
Awesome work! Just take a look at Santi's feedback and resolve them, he points out some naming and formatting things I agree with. I think this would be nice to merge before the next playtests this afternoon after you got the new double jump in!

engine/bindings/private/physics/physics_bindings.cpp
dumitrurares14: because the new dash system interolates the position of the body
```js
 playerBody.SetTranslation(Math.Mix(playerBody.GetPosition(), dashWishPosition, 0.1))
 ```

File: engine/physics/public/components/rigidbody_component.hpp (hunk 4 above)
Comment (dumitrurares14) [reply engine/physics/public/components/rigidbody_component.hpp:35]: During the dash.
Since the whole movement is physics based I am only playing around with the rigid body. So far I only added velocities but with the new dash system I modify the position directly so I need this

File: engine/bindings/private/physics/physics_bindings.cpp (hunk 2 above)
Comment (dumitrurares14) [reply engine/bindings/private/physics/physics_bindings.cpp:93]: I have no idea how to do it differently. I just followed our current implementation. The normal and hit position is implemented the same for the hit info, always has been

```cpp
    rayHitInfo.propReadonlyExt<bindings::GetRayHitPosition>("position");
    rayHitInfo.propReadonlyExt<bindings::GetRayHitNormal>("normal");
    rayHitInfo.propReadonlyExt<bindings::GetRayHitFraction>("hitFraction");
```

game/gameplay/movement.wren
dumitrurares14: Fixed

game/gameplay/movement.wren
dumitrurares14: Fixed

engine/bindings/private/physics/physics_bindings.cpp
santiberna: If there is no trivial alternative then I accept it, although this is kind of a code smell since now physics bodies have two ways to set their position (setting the transform and setting the translation in the rigidbody)

engine/bindings/private/physics/physics_bindings.cpp
santiberna: Check wrenbind docs for how to do this, since it avoids having to create these getter functions.

engine/bindings/private/physics/physics_bindings.cpp
dumitrurares14: I am not really sure if altering dirreclty the transform component would work in this scenario. I think i could waste quite some time if I try to rewrite it. But if you think about it it kinda makes sense also since movement is physics bound, velocity changes the position with jolt, me calling that function is still changing position with jolt but not physically based. I hope this makes more sense now

engine/bindings/private/utility/math_bind.hpp
dumitrurares14: Fixed

engine/bindings/private/physics/physics_bindings.cpp
dumitrurares14: fixed

```cpp
rayHitInfo.var<&RayHitInfo::position>("position");
    rayHitInfo.var<&RayHitInfo::normal>("normal");
    rayHitInfo.var<&RayHitInfo::hitFraction>("hitFraction");
    
    ```

santiberna (approved):
Yup, feedback was implemented. Approved

PR #198, opened 2025-03-20
Refactor Spree: Move all resource code to model loading

santiberna:
### Reviewer steps

### Description

This PR moves all resource code to the model loading module for easier possible improvements in the future

### Test criteria

Feedback on this refactor is needed on:
- Whether all resources should belong in the model loading module (Buffer might not fit)
- Whether we can rename model loading into resources module
- Whether resource managers also belong in this module


MuskettaMan (approved):
It looks good to me, but I would rename the module to resources instead. A modelLoading module implies it exists for the purpose of loading modules, while now the renderer also needs to be dependent on this module even if it doesnt want to load models.

mmzala (approved):

PR #203, opened 2025-03-25
Feature/analytics bindings

MuskettaMan:
### Reviewer steps

### Description

Implement bindings into game analytics SDK

### Issues

- https://bubonic-brotherhood.codecks.io/card/1rq-create-analytics-script-bindings


mmzala (approved):
LGTM

game/analytics/analytics.wren
```diff
@@ -0,0 +1,5 @@
+import "engine_api.wren" for Analytics
+
+class AnalyticsManager {
```
santiberna, line 3: I agree with this, but we need to make sure that the rest of the game logic is well abstracted so we do not need to pass this class everywhere to send analytics events

game/game.wren
```diff
@@ -3,15 +3,13 @@ import "gameplay/movement.wren" for PlayerMovement
...
 class Main {
 
     static Start(engine) {
-
-
         engine.GetGame().SetHUDEnabled(true)
-        
-        
+
         // Set navigational mesh
```
santiberna, line 13: Can we maybe already get an analytics event firing on game start, like a "Start Game"?

engine/analytics/private/analytics_module.cpp
```diff
@@ -47,8 +47,6 @@ ModuleTickOrder AnalyticsModule::Init(MAYBE_UNUSED Engine& engine)
         gameanalytics::GameAnalytics::setEnabledEventSubmission(true); // TODO: Look into privacy policy later for distribution
```
santiberna, line 47: Is it not important that we look into this now before we upload to Steam? (we can possibly disable it for distribution before we get a privacy policy screen)

santiberna (changes requested):
In general everything looks good to me, I just have some questions that should be discussed. Also we have to merge from main

luuk234 (approved):
I don't see any problems here, although some common events would be nice, like game start, game end but that can be done in a later pull request

game/analytics/analytics.wren
MuskettaMan: No, the idea is this will have static methods that directly relate to the analytics documents. So like `WeaponAccuracyEvent(weapon, accuracy)`, so it becomes a simplification on top of the analytics

engine/analytics/private/analytics_module.cpp
luuk234: this

game/game.wren
MuskettaMan: It already does initialize events, it doesn't do progression events yet, but it doesnt make sense until we're able to complete them too

game/analytics/analytics.wren
santiberna: Oh I see... okay that also works

game/game.wren
santiberna: Okay this can wait until gameplay for another PR

engine/analytics/private/analytics_module.cpp
santiberna: Yeah seems like the best bet

santiberna (approved):
Questions were answered. Approved!

PR #224, opened 2025-04-01
Feature/foreground rendering

mmzala:
### Reviewer steps

### Description

Adds support for rendering objects in the foreground.

foreground

### Issues

[codecks task](https://bubonic-brotherhood.codecks.io/milestones/run/65/card/1ug-rendering-on-the-foreground)

### Test criteria

- [x] The gun is always rendered in the foreground


engine/bindings/private/entity/entity_bind.cpp
```diff
@@ -193,6 +194,28 @@ void DetachChild(WrenEntity& self, WrenEntity& child)
     RelationshipHelpers::DetachChild(*self.registry, self.entity, child.entity);
 }
 
+void AddRenderInForegroundRecursive(entt::registry& registry, entt::entity entity)
+{
+    registry.emplace<RenderInForeground>(entity);
+
```
LeoHellmig, line 200: If this function is called on the same entity twice the application will crash.
Use Use emplace_or_replace instead

engine/renderer/private/gpu_scene.cpp
```diff
@@ -197,21 +199,34 @@ void GPUScene::UpdateObjectInstancesData(uint32_t frameIndex)
...
+            _staticDrawCommands.emplace_back(DrawIndexedIndirectCommand {
+                .command = {
+                    .indexCount = mesh->count,
+                    .instanceCount = 0,
+                    .firstIndex = mesh->indexOffset,
+                    .vertexOffset = static_cast<int32_t>(mesh->vertexOffset),
+                    .firstInstance = 0,
+                },
+            });
+        }
```
LeoHellmig, line 222: In the staticMeshView, RenderInForeground is excluded, so there is no reason to check for it

LeoHellmig: Or maybe this wont render static meshes in the foreground at all

LeoHellmig (changes requested):
I found 2 possible issues with using EnTT
The feature works for the gun and looks cool

engine/renderer/private/gpu_scene.cpp
mmzala: Ah, you're right. I forgot to remove that

PR #230, opened 2025-04-01
[FIX] Camera yaw locked

LarsJanssen21:
### Reviewer steps

### Description

Changed camera system to rely on pitch and yaw values to determine the forward vector and clamp the camera to avoid flipping the horizon

ClampedCamera

### Issues

https://bubonic-brotherhood.codecks.io/milestones/run/65/card/1e3-clamping-camera-no-longer-works

### Test criteria

Moving the camera around should behave the exact same way as before, save for the change that the camera yaw is now clamped to prevent flipping backwards


luuk234 (approved):
nothng much here, looks good, builds

PR #258, opened 2025-05-14
Feature/test emitter + serialization

lyvanettinger:
### Reviewer steps

### Description

- Fixed emitter not spawning next to the player when spawning from the particle editor, making it harder to find them
- Added a tag to emitters spawned from the particle editor to update its variables from its preset every frame, making them live editable through their preset
- Moved all emitter presets to be serialized from file

### Issues

https://bubonic-brotherhood.codecks.io/card/1xv-editor-test-particles
https://bubonic-brotherhood.codecks.io/card/1gs-serialize-emitter-presets

### Test criteria

- [x] Spawn a particle preset from the particle editor window and edit its variables in the particle editor window
- [x] Edit a preset, click the save preset button and check if it loads the edited preset when re-opening the application


dumitrurares14 (approved):
Looks good to me. I would suggest having each preset as a separate file so you could save/ replace them individually, otherwise you are always forced to save all of them and maybe you'd want to keep changes only to one. It's not a big deal but its a design choice worth considering.

santiberna (commented):
Some nitpicks, but otherwise everything has been tested. Relating to a file per emitter, it is more elegant but requires a file picker dialog (and also barely changes anything from the user perspective). I do think we could move the emitter presets somewhere else (same with settings.json)

engine/inspector/private/menus/particle_editor.cpp
```diff
@@ -226,13 +231,18 @@ void ParticleEditor::RenderEmitterPresetEditor()
 
             TransformComponent transform;
             _ecsModule.GetRegistry().emplace<TransformComponent>(entity, transform);
-            const auto view = _ecsModule.GetRegistry().view<CameraComponent, TransformComponent>();
-            for (const auto cameraEntity : view)
+            const auto view = _ecsModule.GetRegistry().view<NameComponent, TransformComponent>();
```
santiberna, line 234: Could also use player tag, removes string checking and all of that

engine/particles/public/emitter_component.hpp
```diff
@@ -1,30 +1,28 @@
...
 #include <entt/entity/registry.hpp>
 
+enum class EmitterPresetID : uint8_t;
+
 struct ActiveEmitterTag
 {
 };
 
-struct ParticleBurst
+struct TestEmitterTag
```
santiberna, line 14: What is this tag exactly used for? Place a comment if necessary so noone deletes it (if it is necessary for testing)

lyvanettinger: This tag makes sure that only the emitters spawned through the editor are live editable. I'll place a comment for it

engine/particles/public/particle_module.hpp
```diff
@@ -52,7 +55,8 @@ class ParticleModule final : public ModuleInterface
     std::string_view GetName() override { return "Particle Module"; }
 
 public:
-    ParticleModule() = default;
+    ParticleModule()
+        : _emitterPresets("emitter_presets.json") {};
```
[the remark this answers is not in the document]
lyvanettinger: I agree and think moving it under /game/ would be nice.

santiberna (approved):
Nice, feedback was implemented. Approved!

PR #262, opened 2025-05-15
Feature/coins

dumitrurares14:
### Reviewer steps

### Description

https://github.com/user-attachments/assets/a1bf09e4-42f7-43b6-a223-e80a52bf8e7c

This PR adds the gold nuggets system to our game where enemies will spawn physics based nuggets when they die.

This PR also changed a bit some physics issue that were forgotten in previous blocks and everything should work nicely now. Keep in mind that the enemies or anything that has a transform and a rigid body too, its not recommended to move it via the rb transform instead of our transform system.

### Issues

https://bubonic-brotherhood.codecks.io/decks/46-gameplay/card/1y2-gold-system

### Test criteria

- [ ] Give it a try during gameplay, and leave some feedback.
- [ ] Let me know if i broke anything with the new physics changes

### DIFF
--- a/engine/bindings/private/utility/random_util.hpp
+++ b/engine/bindings/private/utility/random_util.hpp
@@ -73,11 +73,6 @@ private:
     static std::uniform_int_distribution<std::mt19937::result_type> dist;
 };
 
-// Define static members outside the class
-std::random_device RandomUtil::dev;
-std::mt19937 RandomUtil::rng(RandomUtil::dev());
-std::uniform_int_distribution<std::mt19937::result_type> RandomUtil::dist(0, std::numeric_limits<uint32_t>::max());
-
 class Perlin
 {
 public:
@@ -94,7 +89,7 @@ public:
     }
 };
 
-Perlin CreatePerlin(uint32_t seed) { return Perlin { seed }; }
+inline Perlin CreatePerlin(uint32_t seed) { return Perlin { seed }; }
 
 inline void BindRandom(wren::ForeignModule& module)
 {
--- a/engine/bindings/private/utility/random_util.hpp
+++ b/engine/bindings/private/utility/random_util.hpp
@@ -15,11 +15,7 @@ class RandomUtil
 public:
     static uint32_t Random()
     {
-        std::random_device dev;
-        std::mt19937 rng(dev());
-        std::uniform_int_distribution<std::mt19937::result_type> dist6(0, std::numeric_limits<uint32_t>::max());
-
-        return dist6(rng);
+        return dist(rng);
     }
 
     static uint32_t RandomIndex(uint32_t start, uint32_t end)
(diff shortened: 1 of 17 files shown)


MuskettaMan (approved):
Coins look good and work well, nice job!

engine/bindings/private/utility/random_util.hpp
```diff
@@ -70,8 +66,18 @@ class RandomUtil
...
         return glm::vec3(x, y, z);
     }
+
+private:
+    static std::random_device dev;
+    static std::mt19937 rng;
+    static std::uniform_int_distribution<std::mt19937::result_type> dist;
 };
 
+// Define static members outside the class
```
LarsJanssen21, line 76: Any reason why this is not in a cpp file? this risks potentially creating 2 or more copies of the same static variable

LarsJanssen21 (changes requested):
Couple of questions that might not necessarily need changes but good to look at nonetheless

engine/bindings/private/utility/random_util.hpp
dumitrurares14: Before they were being created, every time you called random. I am not sure how bad that would affect performance, but I am sure the random distribution was messed up because of that. So, since we still want this to be static and be called without an object, I moved them from Random and made them static in the class.

engine/bindings/private/wren_bindings.cpp
```diff
@@ -41,12 +41,12 @@ void TransitionToScript(WrenEngine& engine, const std::string& path)
     engine.instance->GetModule<GameModule>().TransitionScene(path);
 }
 
-WrenEntity LoadModelScripting(WrenEngine& engine, const std::string& path)
+WrenEntity LoadModelScripting(WrenEngine& engine, const std::string& path, bool hasPhysics)
```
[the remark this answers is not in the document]
dumitrurares14: The only thing this changes is to have control over whether a loaded object without animations will have a convex collision generated and placed into the ECS as a rigidbody or not. It's just a bool that's being passed through the system, and in the end it's used for an if statement to check if we should create a body or not. This was always an issue with our engine since everything we loaded that was static mesh was loaded with collision too, and we might not always want that (for example, the nuggets, I want them to be simple shapes for performance and memory, cubes, not convex shapes). 

I am not completely sure what the question was tho, I might have misunderstood it xD

engine/bindings/private/utility/random_util.hpp
LarsJanssen21: I get the point, but everytime this hpp file gets used in a cpp file it creates a new copy of the static member variables since this will never be exposed to the linker

engine/bindings/private/wren_bindings.cpp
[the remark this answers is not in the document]
dumitrurares14: Yes indeed, this is just a parameter that will decide if the model will have collision on it or not

engine/bindings/private/utility/random_util.hpp
dumitrurares14: Ight, I moved to cpp file to avoid linker buba

engine/bindings/private/wren_bindings.cpp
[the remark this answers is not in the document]
LarsJanssen21: I like the name!

LarsJanssen21 (approved):
Awesome work!

PR #281, opened 2025-05-26
Fix/reset cleanup

santiberna:
### Reviewer steps

### Description

This PR fixes a bunch of issues that occurred when something from the previous scene "leaked" into the next one (because wren shutdown code missed something or an error occurred and shutdown was not able to run). The fixes:
- Audio Module is reset after a scene transition
- ECS registry is reset after a scene transition
- Decals are cleared
- Particles are cleared
- Flash colour is reset

### Issues

https://bubonic-brotherhood.codecks.io/card/1xy-sounds-and-music-keep-playing-after-changing-scenes

### Test criteria

- [ ] Trigger a long sound effect (like the eye charging)  and go to the main menu immediately (check if it still continues playing)
- [ ] Trigger a red health flash from getting hit by a beserker, then pause and go to main menu (check if there is still a red flash)


dumitrurares14 (changes requested):
Hitting the reload button in engine breaks some stuff. I don't think hot reloading is working properly now. We get the "only one light is supported" warning spam in console and i am seeing the revolver from before reload in the new scene.

PR #292, opened 2025-05-28
Fix/physics transform checks

santiberna:
### Reviewer steps

### Description

WARNING: if we want this before release, it must be thoroughly tested since it changes the way physics works in the engine.

This PR removes some unnecessary parts of the physics / transform synchronization to make it easier to debug. It also includes an assert when the transform of a rigidbody entity is modified, since it is not the intended way of manipulating physics objects. When this happens, THE RIGID BODY WILL ALWAYS TAKE PRIORITY and as a result changing the transform does not do anything.

I took the liberty of also fixing all the instances where scripting with physics objects modifies transforms.

### Test criteria

- [x] The game must function exactly as before!
- [x] Check if no asserts are hit (related to modifying physics bodies through transform)


LeoHellmig (changes requested):
The game crashes relating to the coins and coins not falling or moving towards the player
The player is not reset to underneath the map when they go out of bounds
The directional light should not have been changed

mmzala (changes requested):
Aside from what Leo already mentioned, the code looks good. The coins are broken though and don't move to you anymore when you're close.

mmzala (approved):
The coins work now, nice work!

LeoHellmig (approved):
Things work 👍

PR #299, opened 2025-06-04
Fix/jolt asserts

santiberna:
### Reviewer steps

### Description

This PR fixes some issues with the way we handle physics meshes. Mesh collision is no longer generated for models that are used with capsule colliders (optimizing a bit).

PULL FROM PERFORCE FOR THE FIXED COLLISION

OLD ~ 50 000 verts
image

NEW ~ 17 000
image

### Test criteria

- [ ] Test moving around the map with the simplified collision mesh
- [x] Check if Jolt Assert is still hit
- [ ] Check if map collision is inaccurate / less consistent


dumitrurares14 (approved):
Nice job !!

PR #310, opened 2025-06-10
Added extra checks to avoid hard crashes from audio module

dumitrurares14:
### Reviewer steps

### Description

This PR adds some extra checks in the Audio Module to avoid hard crashes during gameplay if certain funcitons are called but the event instances are deleted meanwhile.

```cpp
// This should never happen but lets avoid a hard crash
if (!FMOD_Studio_EventInstance_IsValid(_events[instance.id]))
    return;
```

### Issues
None

### Test criteria

None

### DIFF
--- a/engine/audio/private/audio_module.cpp
+++ b/engine/audio/private/audio_module.cpp
@@ -83,6 +83,9 @@ void AudioModule::Reset()
         if (event == nullptr)
             continue;
 
+        if (!FMOD_Studio_EventInstance_IsValid(event))
+            continue;
+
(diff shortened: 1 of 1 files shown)


LeoHellmig (approved):
Should have been in there a long time ago tbh, awesome

santiberna (changes requested):
Could we also add these checks on the freeing of all sounds under ``AudioModule::Reset()``?

I also have been encountering this error when shutting down FMOD, might be relevant to fix in this PR.

`[FMOD ERROR] : 844 ( closeInternal ) - assertion: 'connectionsRemaining == 0' failed`

santiberna (approved):
Fixed DSP free!

PR #311, opened 2025-06-10
Improved skeleton movement

LarsJanssen21:
### What
- Skeletons stop charging when the altitude difference is to high and resort back to pathfinding to player position
- Skeletons avoid local objects when running towards the player
- Skeletons that are using a path to guide their way to the player use a improved heuristic that forces movement in a straighter line

ObstacleAvoidance

### What to check for
Some things are a bit subjective but in general:
- Playing the game should feel more challenging with enemies that are directed more towards the player
- Skeletons shouldn't get stuck on local obstacles (such as gravestones, houses, lampposts and the like)
- Generally investigate whether the improved skeletons make for a better feeling game
- Check if enemies are not getting stuck preventing you from smoothly transitioning to the next round

**Don't forget to pull from perforce, otherwise the PR won't work**


engine/pathfinding/private/pathfinding_module.cpp
```diff
@@ -316,7 +316,7 @@ float PathfindingModule::DirectedHeuristic(glm::vec3 startPos, glm::vec3 endPos,
 
     float dot = std::max(glm::dot(towardsEnd, towardsFinal), 0.001f);
 
-    return (1.0f - dot) * towardsEndLength;
+    return powf((1.0f - dot), 3.0f) * towardsEndLength;
```
santiberna, line 319: Sometimes the enemies take slightly longer paths towards the player (like they are strafing). Is this because of heuristics or a navmesh issue?

santiberna (changes requested):
In general, the enemies seem to not get stuck anymore. However, I have noticed a couple problems with the map collision that make this a bit hard to test:

- The broken bridge area has missing collision on the floor
- There are a couple invisible walls around the level
- You can just go out of bounds into the ocean (the enemies also followed me a little bit there)

image

In terms of the code, the local steering does seem to work well and the implementation is clean. Kudos for also cleaning up the enemy logic a bit. There is still one problem: you did not implement this for the berserker.

It is not strictly necessary to get a fixed navmesh and collision to merge this PR, but atleast address these concerns and implement this for the berserker

engine/pathfinding/private/pathfinding_module.cpp
LarsJanssen21: I honestly haven't really delved too deep into what causes the sometimes illogical paths. But since most of the time during gameplay it's not that noticeable I haven't made it that high of a priority.

With this new heuristic it should more heavily favor straight paths which it does seem to do, with some outliers here and there

LarsJanssen21:
> In general, the enemies seem to not get stuck anymore. However, I have noticed a couple problems with the map collision that make this a bit hard to test:
> 
> * The broken bridge area has missing collision on the floor
> * There are a couple invisible walls around the level
> * You can just go out of bounds into the ocean (the enemies also followed me a little bit there)
> 
> image
> 
> In terms of the code, the local steering does seem to work well and the implementation is clean. Kudos for also cleaning up the enemy logic a bit. There is still one problem: you did not implement this for the berserker.
> 
> It is not strictly necessary to get a fixed navmesh and collision to merge this PR, but atleast address these concerns and implement this for the berserker

As for the berserker this will be done in a separate PR, but you're right in saying that I should have mentioned this!

PR #314, opened 2025-06-11
[Fix] TransformHelpers::GetWorldRotation()

santiberna:
### Description

Fixes a bug in the aforementioned function


dumitrurares14 (approved):
I tested and the gold and map collision are working.

